PR #40, opened 2026-05-08
test(LinkField): add null-safety regression suite + export LinkOptionEntity

ankitDhwani:
| Point | Verdict | Action |
|---|---|---|
| No widget tests (blocking) | ✅ Valid | Written — see below |
| FormBuilderDropdown<String> + value: null type risk (blocking) | ⚠️ Overcautious — code is safe with `flutter_form_builder: 10.3.0+2` (`T?` is nullable by design) | No change needed; test suite confirms runtime safety |
| PR description thin (minor) | ✅ Valid | Out of scope for code change; note it for future PRs |

## Code changes

- `lib/frappe_mobile_sdk.dart` — exports `LinkOptionEntity`. It was already the return type of `LinkOptionService.getLinkOptions` but was unreachable to SDK consumers building custom link services.

### Loading state
Asserts on `formKey.currentState!.value['test_link']` (the form's serialized output):

- Form value is `null`, never `'__blank__'`, when there's no prior value
- No sentinel string in any form value map entry during loading
- Prior value is displayed visually; no sentinel in any `DropdownMenuItem`

### Empty state
The exact path that burned Devang:

- Form value is `null` (not `'__blank__'`) when options list is empty
- No `DropdownMenuItem.value == '__blank__'` anywhere in the tree
- `onChanged` emits `null` (not `'__blank__'`) when the hint item is tapped — verified via UI interaction + captured callback

### Waiting-for-dependent state

- `"Select state first"` hint text shown, refresh icon absent
- Form value is `null` while waiting
- Options load normally when the dependent field value is present

### Options loaded

- `SearchableSelect` renders after service resolves; no sentinel text in widget tree
- Single-option auto-select fires `onChanged` with the correct value
- Clear fires `onChanged(null)` — confirmed `null`, not a sentinel

### DIFF
--- a/lib/frappe_mobile_sdk.dart
+++ b/lib/frappe_mobile_sdk.dart
@@ -20,6 +20,7 @@ export 'src/models/workflow_transition.dart';
 export 'src/database/app_database.dart';
 export 'src/database/entities/doctype_meta_entity.dart';
 export 'src/database/entities/document_entity.dart';
+export 'src/database/entities/link_option_entity.dart';
 export 'src/database/daos/doctype_meta_dao.dart';
 export 'src/database/daos/document_dao.dart';
 
--- a/test/link_field_test.dart
+++ b/test/link_field_test.dart
@@ -12,7 +12,7 @@ import 'package:flutter_form_builder/flutter_form_builder.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:frappe_mobile_sdk/frappe_mobile_sdk.dart';
 import 'package:frappe_mobile_sdk/src/ui/widgets/fields/searchable_select.dart';
-import 'package:sqflite_common_ffi/sqflite_ffi.dart';
+
 
 // ---------------------------------------------------------------------------
 // Test doubles
@@ -76,10 +76,7 @@ LinkOptionEntity _opt(String name) => LinkOptionEntity(
 // ---------------------------------------------------------------------------
 
 void main() {
-  setUpAll(() {
-    sqfliteFfiInit();
-    databaseFactory = databaseFactoryFfi;
-  });
+
 
   // ---- Loading state -------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


deepak-dhwani:
Good structure overall — four state groups, deterministic async via Completer, and reason: on every assertion. The LinkOptionEntity export is the right call. Three things before merge:                      
                                                                                                                                                                                                                
  ISSUE-1 — Remove the setUpAll SQLite block (lines 77–80)                                                                                                                                                      
                                                                                                                                                                                                                
  None of these tests touch the DB. sqfliteFfiInit() + databaseFactory = databaseFactoryFfi is dead weight here, adds a spurious dev-dependency, and misleads future maintainers into thinking DB access is     
  involved. Drop the block and the sqflite_common_ffi import entirely. If tests fail without it, that points to a production code design problem — not something to paper over in test setup.                   
                                                                                                                                                                                                                
  ISSUE-2 — Test 12: don't call the widget's internal callback directly                                                                                                                                         
   
  tester.widget<SearchableSelect>(searchableSelect).onChanged([]);                                                                                                                                              
                                                                                                                                                                                                                
  This bypasses the actual user interaction and couples the test to SearchableSelect's API shape. Prefer tapping the real clear affordance (find.byIcon(Icons.clear) inside SearchableSelect). If no clear      
  button exists, add a comment explaining why the callback is invoked directly — otherwise it reads as a shortcut.                                                                                              
                                                                                                                                                                                                                
  ISSUE-3 — Add a didUpdateWidget transition test                                                                                                                                                               
   
  The three waiting-for-dependent tests only cover static formData. There's no test that rebuilds the widget with a new formData (simulating a dependent field being filled in) and verifies the transition:    
  waiting → loading → SearchableSelect rendered. This is the exact code path the original bug lived in — didUpdateWidget → _loadOptions. Without it, a regression in that path would not be caught by this
  suite.                                                                                                                                                                                                        
                  
  testWidgets('options load when dependent field value is filled in', (tester) async {
    // render with empty formData → waiting state                                                                                                                                                               
    // rebuild with {'state': 'Meghalaya'} → should exit waiting, enter loading                                                                                                                                 
    // resolve options → SearchableSelect appears, form value stays null                                                                                                                                        
  });

deepak-dhwani (approved):

PR #69, opened 2026-06-02
feat(utils): export DependsOnEvaluator on the public barrel

abhijitnairDhwani:
## What

One-line addition to the SDK barrel:

```dart
export 'src/utils/depends_on_evaluator.dart' show DependsOnEvaluator;
```

No code in `src/utils/depends_on_evaluator.dart` changes; only its visibility.

## Why this is needed

`DependsOnEvaluator` evaluates Frappe's `depends_on` / `mandatory_depends_on` expressions. The class is already mature canonical infrastructure for the SDK itself — referenced **9 times internally**:

- `src/ui/widgets/form_builder.dart` × 7 (visibility, mandatoriness, read-only, tab/section evaluation)
- `src/services/link_option_service.dart` × 2 (`extractEvalDocField`)

The class docstring already calls out the public-API intent:

> **Trust boundary:** expressions evaluated here come from server-side DocType meta configured by Frappe admins, not from end-user input. This evaluator is NOT sandboxed — do not pass user-supplied strings to `evaluate`.

The class lives in `src/utils/` only because it was never explicitly exported. Consumers reaching into the SDK to use it need `// ignore: implementation_imports`.

## Who needs this

Apps that **own their own form layer** rather than rendering through `FrappeFormBuilder` need the same evaluator on the UI side. Concretely: any consumer that:

1. Uses the SDK's data layer (`OfflineRepository`, `SyncService`, `UnifiedResolver`) for offline-first save / pull / push;
2. But builds their own widget tree for forms (because of design-system requirements, custom layouts driven by client-locked design specs, etc.);

… needs `depends_on` / `mandatory_depends_on` semantics on the UI side to match the SDK's data-layer expectations.

Re-implementing the evaluator locally would mean duplicating ~150 LOC of expression parsing (the `eval:` prefix strip, AND/OR splits outside brackets, six comparison operators, `.includes([...])` patterns, nested-paren handling). Every time the SDK extends the supported expression syntax, the consumer's duplicate would silently drift.

Exporting the symbol from the SDK's public barrel keeps semantics aligned with the upstream evaluator, with no code drift surface.

## Why not introduce a new abstract interface

The existing class is the contract. There's nothing implementation-specific about evaluating a string against a `Map<String, dynamic>` — that's exactly what the SDK does internally too. A new interface would just shadow the existing class without adding value.

## Tests

No new tests; the change is visibility-only. The existing evaluator tests in `test/utils/depends_on_evaluator_test.dart` (if any) and the 7+2 internal call sites continue to work unchanged.

### DIFF
--- a/lib/frappe_mobile_sdk.dart
+++ b/lib/frappe_mobile_sdk.dart
@@ -135,3 +135,4 @@ export 'src/ui/dialogs/force_logout_confirm.dart' show showForceLogoutConfirm;
 // Utils (debug tracer + user-friendly errors)
 export 'src/api/utils.dart' show extractErrorMessage, toUserFriendlyMessage;
 export 'src/utils/api_tracer.dart' show ApiTracer;
+export 'src/utils/depends_on_evaluator.dart' show DependsOnEvaluator;


Omprakash-48 (approved):

deepak-dhwani (changes requested):
PR #69 — feat(utils): export DependsOnEvaluator on the public barrel — Round 1 Review
                                                                                                                                               
  One line added to lib/frappe_mobile_sdk.dart. The substance of the review is whether the class being exposed is ready for a public API contract.

  ---
  The change

  export 'src/utils/depends_on_evaluator.dart' show DependsOnEvaluator;

  show DependsOnEvaluator exports the entire class, which means two public static methods land on the SDK surface:

  - DependsOnEvaluator.evaluate(String? expression, Map<String, dynamic> formData) — the main evaluator, clearly intended for consumers
  - DependsOnEvaluator.extractEvalDocField(String value) — a narrower utility currently used internally in link_option_service.dart, now also public

  Both are purely functional with no instance state, which makes the public contract clean.

  ---
  Correctness review of the exported implementation
  
  The evaluator logic is sound for its documented scope.

  Operator precedence is correct — && is split before ||, matching standard precedence. The bracket-aware splitting in _splitOutsideBrackets correctly handles
  .includes([...]) arrays spanning &&/|| delimiters.

  Operator substring ordering is correct — === is checked before ==, !== before !=, >= before >, <= before <. All the substring-collision hazards are handled.

  _stripOuterParens is correct — it only strips parens that wrap the entire expression, so (A) || (B) is preserved.

  extractEvalDocField works correctly — the expr == fieldName sentinel for "not a doc.field expression" is correct. eval:doc.x returns "x", eval:something_static
  returns null.

  ---
  MEDIUM M1 — Test file still imports from the internal path

  test/depends_on_evaluator_test.dart line 2

  import 'package:frappe_mobile_sdk/src/utils/depends_on_evaluator.dart';

  Now that the class is exported on the public barrel, this should use the public import:

  import 'package:frappe_mobile_sdk/frappe_mobile_sdk.dart' show DependsOnEvaluator;

  The practical value: if the barrel export is ever accidentally removed or the show clause is narrowed, the test continues to pass against the internal path and
  gives no signal. Updating the import makes the test guard the actual public surface.
  
  ---
  MEDIUM M2 — No tests for ===, !==, or any numeric operator
  
  The evaluator implements six operators beyond == and !=: ===, !==, >, <, >=, <=. None have tests. These are all exercised by real Frappe DocType expressions —
  Frappe desktop frequently uses === for strict comparisons and numeric thresholds like doc.age >= 18. Now that this is public API, consumers will use it against
  expressions in the wild and expect these operators to work. The existing suite would not catch a regression in any of them.

  At minimum one test per operator is needed:                                                                                                  

  test('=== strict equality', () {
    expect(DependsOnEvaluator.evaluate("eval:doc.x === 'Y'", {'x': 'Y'}), isTrue);
    expect(DependsOnEvaluator.evaluate("eval:doc.x === 'Y'", {'x': 'N'}), isFalse);
  });
  test('>= numeric threshold', () {
    expect(DependsOnEvaluator.evaluate('eval:doc.age >= 18', {'age': 18}), isTrue);
    expect(DependsOnEvaluator.evaluate('eval:doc.age >= 18', {'age': 17}), isFalse);
  });
  // similarly for !==, >, <, <=

  ---
  LOW L1 — extractEvalDocField has no doc comment
  
  The method is now public API. Its contract is non-obvious: it returns null both when the input has no eval:doc. pattern and when the expression is not a
  doc-reference at all. A one-line comment explaining the return value and the null sentinel would save every consumer from reading the implementation.

  ---
  LOW L2 — Error default not documented for mandatory_depends_on callers
  
  The method doc mentions both depends_on and mandatory_depends_on but does not say that a parse failure returns true. For depends_on that means show the field —
  safe. For mandatory_depends_on it means make the field mandatory — potentially surprising. Worth a note in the doc comment: "Returns true on any parse error —
  callers using this for mandatory_depends_on should be aware that an unrecognised expression will make the field mandatory."

  ---
  Verdict
  
  Not blocking on the export decision — the class is ready for public use. M1 and M2 should be addressed before merge: the test import gap means the barrel export
  is not actually guarded by the test suite, and the missing operator tests leave the public API unguarded against regression on common Frappe numeric expressions.

PR #78, opened 2026-06-18
Feat/sdk mobile error log

Omprakash-48:


bhushan-barbuddhe (approved):

deepak-dhwani (commented):
PR #78 — Feat/sdk mobile error log — Round 1 Review
                                                                                                                                               
  This PR introduces three significant changes: a mobile error log pipeline that captures terminal push failures and posts them to the server, a frappeToPushError
  adapter that fixes a long-standing bug where all server rejections were classified as UNKNOWN, and an isInitialSync bulk-insert optimisation in PullApply. The
  core architecture is well thought out, but there are two blockers and several highs that need to go before this can merge.
  
  ---
  BLOCKER B1 — Full document payload (PII) is stored and posted to the server in error logs
  
  lib/src/services/error_capture.dart, lib/src/sync/mobile_error_record.dart, lib/src/api/exceptions.dart

  MobileErrorRecord.requestPayload stores the entire HTTP request body:

  requestPayload: error.requestBody != null
      ? jsonEncode(error.requestBody)
      : jsonEncode(payload),

  For a POST or PUT operation, error.requestBody is the full Frappe document map — patient name, date of birth, address, GPS coordinates, medical record data,
  custom form responses. This is serialised verbatim and posted to mobile_sync.report_error, where it is stored in a Frappe DocType visible from the desk. A
  single compromised desk account can then read patient data for every sync failure across the deployment.
  
  India's DPDP Act 2023 classifies health data as sensitive personal data. Storing it in an error-log table with a different purpose basis from the original
  collection is a compliance failure regardless of whether the endpoint is access-controlled.

  The FrappeException.requestBody field (which feeds requestPayload) is also Object? stored on the exception object itself. If Crashlytics or Sentry serialises
  the exception, or if toString() is ever updated to include these fields, the document data leaks into crash reporting infrastructure.

  Fix: replace requestPayload with a structural skeleton — field names and types, not values. A helper that deep-nulls all leaf values preserves debugging value
  (which fields were present, which types) without transmitting any patient data. The requestBody field on FrappeException should similarly store only structural
  metadata, not actual values, and should carry a comment explicitly prohibiting it from being passed to crash reporters.
  
  ---
  BLOCKER B2 — HTTP 401 (session expiry) is classified as a terminal PermissionError, permanently pausing all queued rows
  
  lib/src/services/sync_engine_builder.dart, dispatchHttpSend

  on AuthException catch (e) {
      onFailure?.call(e, method, payload);
      throw frappeToPushError(
          status: e.statusCode ?? 403,
          excType: 'PermissionError',   // hardcoded — applies to both 401 AND 403
          ...
      );
  }
  
  frappeToPushError maps PermissionError → ServerRejection → PERMISSION_DENIED → isTerminal = true → markPaused. This is correct for HTTP 403 (the user lacks
  permission). It is wrong for HTTP 401.

  A 401 is a session expiry. The RestHelper already attempts a Bearer token refresh on 401; if that fails, the AuthException(401) propagates to dispatchHttpSend.
  At that point, every pending outbox row in the drain is permanently paused. After the user re-authenticates, those rows stay paused forever and require manual
  retry row by row. In practice, any session timeout mid-sync silently kills the entire queue.
  
  Fix: distinguish on e.statusCode. A 401 should throw something retryable (a NetworkError or a new SessionExpiredError) rather than produce a terminal
  classification. For session-cookie auth (no refresh mechanism), the same applies — a 401 means "log in again and retry," not "you will never have permission."

  ---
  HIGH H1 — ErrorLogCollector._bySig has no size cap — unbounded memory and POST count

  lib/src/sync/error_log_collector.dart

  _bySig is a plain Map<String, _Agg> with no eviction policy. In a degraded-connectivity session with a large queue and diverse error types, the map can grow to
  hundreds of entries, each holding up to five ErrorExample objects with the full request payload string. The downstream effect is that MobileErrorPoster.flush
  fires one HTTP call per signature sequentially — N signatures produces N sequential background POSTs with no upper bound.
  
  Fix: cap _bySig at a constant (50 signatures is reasonable) using a LinkedHashMap for O(1) oldest-entry eviction. On the flush side, consider batching all
  aggregated records into a single POST { errors: [...] } rather than one call per signature.

  ---
  HIGH H2 — drain() has a subtle record-loss window

  lib/src/sync/error_log_collector.dart

  List<AggregatedError> drain() {                                                                                                              
      final out = <AggregatedError>[];
      for (final entry in _bySig.entries) { ... }
      _bySig.clear();
      return out;
  }

  The iterate-then-clear pattern means any records added between the start of the for-loop and _bySig.clear() across an async suspension are silently dropped. In
  the current wiring drain() is called synchronously and no suspension happens between the iterate and the clear, so this is safe today. But it is a fragile
  invariant — any future refactor that awaits inside drain() breaks it silently.
  
  Fix: snapshot-and-clear atomically:

  List<AggregatedError> drain() {
      final snapshot = Map<String, _Agg>.of(_bySig);
      _bySig.clear();
      return snapshot.entries.map(...).toList();
  }

  Records added after _bySig.clear() land in the next drain cycle.
  
  ---
  HIGH H3 — 409 with no exc_type defaults to TimestampMismatchError — infinite retry on duplicate entries

  lib/src/services/sync_engine_builder.dart, frappeToPushError

  if (excType == null && status == 409) return TimestampMismatchError();

  HTTP 409 from Frappe is used for multiple conditions. TimestampMismatchError is the most common but not the only one — duplicate name conflicts and custom-app
  conflicts also return 409, and not all of them include exc_type in the response body. When a server returns 409 without exc_type for a non-timestamp reason, the
  adapter maps it to TimestampMismatchError, triggering a timestamp refresh and retry. The retry hits the same 409 again and loops until the retry budget
  exhausts. The row then fails with TIMESTAMP_MISMATCH rather than the actual cause.                                                           

  Fix: default to ServerRejection(status: 409, rawBody: rawBody) for unknown 409s. If the product requirement is genuinely "retry all 409s without exc_type as
  timestamp mismatches," document that contract explicitly in a comment and add a test covering the retry-loop scenario.

  ---
  HIGH H4 — LinkExistsError(linked: const {}) discards the linked-document list from the response body
  
  lib/src/services/sync_engine_builder.dart, frappeToPushError

  case 'LinkExistsError': return LinkExistsError(linked: const {});

  LinkExistsError is supposed to carry the set of documents that prevent deletion — this is what Frappe includes in the error response to tell the user "you can't
  delete this record because X and Y reference it." rawBody is available inside frappeToPushError but is not parsed to extract this information. The result is a
  LinkExistsError with an empty set, so the UI shows "link exists" with no actionable detail. Fix: parse rawBody for the linked-document list, or at minimum pass
  rawBody through so the UI layer can render it.                                                                                               

  ---
  HIGH H5 — StateError for unknown method is not a PushError subtype

  lib/src/services/sync_engine_builder.dart, dispatchHttpSend

  default: throw StateError('dispatchHttpSend: unknown method "$method"');

  StateError is not a PushError subtype. It bypasses onFailure (no error log entry), and whether the engine's outer catch handles it depends on whether that catch
  is typed to PushError or Object. If it only catches PushError, the StateError propagates up unhandled. Fix: throw a structured PushError subtype for invalid
  methods so the engine handles it uniformly and the failure is observable.
  
  ---
  MEDIUM M1 — NetworkException bypasses onFailure — inconsistent observability
  
  dispatchHttpSend calls onFailure for ApiException, ValidationException, and AuthException before rethrowing. It does not call it for NetworkException. Network
  errors are out of scope for the error log per spec §6, which is correct. But the inconsistency is silent — add a comment on the NetworkException branch
  explicitly noting this is intentional per spec §6, so a future reader doesn't add onFailure there by symmetry.

  ---
  MEDIUM M2 — computeSignature includes errorUser — same error across users creates separate server rows

  lib/src/sync/error_log_collector.dart

  Including errorUser in the structural key means the same broken endpoint generates N separate server rows for N field workers — each with occurrenceCount: 1
  instead of one row with occurrenceCount: N. The _Agg.errorUser field already tracks the most-recent user per signature, so per-user attribution is preserved
  without it being part of the dedup key. If including errorUser is intentional (for per-user error tracking), add a comment making that explicit — the test
  'different user => different signature => separate rows' reads more like a description of current behaviour than a deliberate design assertion.

  ---
  MEDIUM M3 — flush() failure silently drops drained records with no re-queue

  lib/src/sync/push_engine.dart

  drain() materialises the collected records into a list and then clears the collector. If flush() subsequently fails (network error), those records are
  permanently lost — the catchError handler logs locally but does not restore them. For "best-effort telemetry" this is a documented design choice per spec §9,
  which is fine. Add a comment on flush() noting that loss-on-failure is intentional, not an oversight, so the next developer does not add retry logic that would
  be counterproductive.                                                                                                                        

  ---
  What's done well
  
  frappeToPushError and dispatchHttpSend are a real correctness fix. Previously all server rejections escaped as ApiException to the engine's catch-all
  markFailed(UNKNOWN), making the timestamp auto-merge, duplicate reconcile, terminal-to-paused, and network-retry paths unreachable in production. The
  integration test suite in push_engine_real_send_error_mapping_test.dart tests the real production dispatch path rather than a mirror, which is exactly the right
  approach — the comment explaining the history ("written RED to prove that, then GREEN") is the kind of documentation that saves future maintainers hours.
  
  The isInitialSync bulk-path optimisation in PullApply is sound. The pre-check COUNT query covers dirty rows, tombstoned rows, and ghost-success rows via the
  mobile_uuid IN (incoming) AND server_name IS NULL clause. The fallback to sequential for any unsafe page is the correct conservative approach. The
  isOwnInsertRoundtrip extension to stamp server_name without overwriting local payload when pending outbox work exists is correct.
  
  The onDrainComplete fire-and-forget wiring is clean: drain() is evaluated synchronously materialising the snapshot before unawaited fires, the double try/catch
  handles both synchronous and async throw modes, and the comment explaining why the hook must not extend the push critical section is helpful.

  recordTerminalFailureSafe is the right pattern — capture must never throw and mask the real sync failure.                                    

  ---
  Verdict

  B1 and B2 need to be resolved before merge. B1 has regulatory implications that make it a hard gate for any healthcare deployment. H1 through H5 should be fixed in the same pass.

PR #80, opened 2026-06-24
Offline-first bug fixes surfaced on a real device, plus option-title translation parity with Frappe web

Omprakash-48:
Offline-first bug fixes surfaced on a real device, plus option-title
translation parity with Frappe web:

- Save crashes when a field is added to a doctype server-side (`no such column`)
- Newly-added fields silently dropped after a mid-session meta refresh
- A Dart-mutex deadlock (hang, no crash) when saving a doc whose meta had to be
  resolved inside the write transaction
- Select/Link dropdowns only tappable in the center — edges of the box were dead
- Link & Table MultiSelect option titles rendered untranslated while Frappe web
  translated them

---

### fix(database): reconcile parent table schema on save

**Symptom:** `SqfliteFfiException: table docs__… has no column named …` on form
submit, after a new field was added to the doctype on the server.

**Root cause:** `OfflineRepository.saveDocument` built its INSERT from the
current `parentMeta.fields` but never reconciled the on-disk `docs__` table
against that meta. Schema reconcile only ran in `ensureSchemaForClosure` (boot,
cached) and `PullEngine` — never on the save path.

**Fix:** `saveDocument` now calls `reconcileParentTableForMeta` before the write
txn (mirrors `PullEngine`'s `schemaReconciler`), ALTER-adding any missing
columns. Runs outside the txn (PRAGMA + ALTER on `rawDatabase`).

### fix(sync): invalidate offline meta cache on server meta refresh

**Symptom:** After a reconnect resync mid-session, saving a record silently
omitted a newly-added field (no crash).

**Root cause:** `OfflineRepository`'s per-doctype meta cache was only cleared on
logout. A mid-session refresh (connectivity listener → `checkAndSyncDoctypes`)
updated the DAO + `MetaService` LRU but left the repository serving a stale
schema.

**Fix:** New `MetaService.onMetaRefreshed` (fired from `_upsertMetaJson`, the
single chokepoint for server meta rewrites) wired in `FrappeSDK` to
`OfflineRepository.invalidateMetaCacheFor`. Local-only — no network added to the
save path.

### feat(link-options): translate Link & Table MultiSelect option titles

**Symptom:** Link and Table MultiSelect option titles rendered untranslated in
the app while Frappe web translated them.

**Root cause:** Both field types fetch options through
`LinkOptionService.getLinkOptions` / `getLinkOptionsOffline`, which built display
labels from synced records but never applied the active translation.

**Fix:** Translate the option **display label** when the field's target doctype
has `translated_doctype` set — matching Frappe core, which gates link/select
title translation on the target doctype (`link.js`, `formatters.js`,
`search.py`). The stored **name** (value pushed to the server) stays English,
same contract as the static Select field. Adds `DocTypeMeta.translatedDoctype`
(mirrors `isSubmittable`), an optional `translate` hook on `LinkOptionService`
gated on `meta.translatedDoctype` in `_rowsToEntities` (covers both field
types), and wires `TranslationService.translate` into `LinkOptionService` at
both `FrappeSDK` construction sites.

### fix(offline): never resolve DocType meta inside the write transaction

**Symptom:** Saving certain docs hung forever — no crash, no `SQLITE_BUSY`, no
timeout.

**Root cause:** `LocalWriter.writeParentInTxn` resolved meta in-txn on two paths
— the parent (`parentMeta ??= _metaResolver(...)`) and the per-child resolver
fallback. `_metaResolver` is `MetaService.getMeta`, which reads the outer
`Database`; invoked inside `rawDatabase.transaction` it queues behind the txn on
sqflite's single non-reentrant lock and deadlocks (a Dart mutex deadlock, never
`SQLITE_BUSY`). `saveDocument` could reach both paths: a null parent meta, or a
child whose meta was not pre-resolved.

**Fix:** `writeParentInTxn` never resolves meta in-txn anymore. `parentMeta` is
now required; a child absent from `childMetasByDoctype` is skipped (loud log)
rather than resolved. `saveDocument` throws a clean `StateError` **before**
opening the txn when the doctype meta is unsynced, and pre-resolves child metas
(skip + loud log when unavailable).

*PR #80 round-1 review follow-ups:* `M1` — extract `_wireMetaRefreshInvalidation()`
called by both `_doInitialize` and `forTesting`, so the `onMetaRefreshed →
invalidateMetaCacheFor` wiring is covered end-to-end. `C` — route
`MetaService.ensureUpToDate`'s schema write through `_upsertMetaJson` so
`onMetaRefreshed` fires. `M2` — document `dropdownFullTap`'s `TextDirection.ltr`
assumption. `L2` — document that `invalidateMetaCacheFor`'s `_childMetasByParent`
eviction is parent-keyed (child-only-refresh staleness is pull-path-only).

### fix(ui): make the full dropdown box open the menu

**Symptom:** Tapping the left/right/top/bottom edges of a Select/Link dropdown
did nothing; only the center responded.

**Root cause:** `DropdownButtonFormField` / `FormBuilderDropdown` only treat the
area inside the `InputDecoration.contentPadding` as clickable — the standard form
style's 16px×12px padding was a dead margin.

**Fix:** New `dropdownFullTap()` helper zeroes contentPadding on all sides and
moves the inset into the dropdown's own (clickable) padding, with
`isExpanded: true`. Value stays visually inset; the whole box opens the menu.
Applied to `SelectField` + the `LinkField` static-options dropdown.

---

## Testing

- **Unit/widget:** full SDK suite **1571 passing**. New regression tests:
  save-path schema heal, mid-session cache invalidation (`onMetaRefreshed` fires;
  stale→fresh) with an end-to-end wiring test, both in-txn deadlock vectors
  (resolver blocks on an outer-DB read → any reintroduced in-txn resolution hangs
  and fails at a 5s timeout), `DocTypeMeta.translatedDoctype` variants,
  `LinkOptionService` translation (translated → label translated / name English;
  not-translated → untouched; no translate fn → no-op), and dropdown edge taps
  (left/center/top/bottom). Red-checks confirmed each test fails without its fix.
- **On-device** (moto e5, Android 8 / API 26): built SNF against this SDK,
  reproduced and confirmed the dropdown full-box tap on the Household Survey form
  (left margin, center, top, bottom all open the menu).

## Files changed

```
lib/src/models/doc_type_meta.dart               # translatedDoctype getter
lib/src/sdk/frappe_sdk.dart                      # wiring: meta-refresh invalidation + translate hook
lib/src/services/link_option_service.dart        # translate option labels
lib/src/services/local_writer.dart               # no in-txn meta resolution
lib/src/services/meta_service.dart               # onMetaRefreshed callback
lib/src/services/offline_repository.dart         # save reconcile + invalidateMetaCacheFor + pre-resolve
lib/src/ui/widgets/fields/field_helpers.dart     # dropdownFullTap()
lib/src/ui/widgets/fields/select_field.dart
lib/src/ui/widgets/fields/link_field.dart
test/models/doc_type_meta_test.dart
test/meta_service_test.dart
test/sdk/frappe_sdk_meta_refresh_wiring_test.dart
test/services/link_option_service_translation_test.dart
test/services/offline_repository_reconcile_test.dart
test/services/offline_repository_local_writer_test.dart
test/services/offline_repository_save_test.dart
test/ui/widgets/fields/select_field_test.dart
```


deepak-dhwani (changes requested):
PR #80 — offline save-path schema/meta correctness + full-box dropdown tap target — Round 1 Review

Three independent fixes, nine files. Each has a concrete reproduction case and a red-check confirmed test. The overall approach is sound.

---
Fix 1 — saveDocument reconciles the parent table schema before writing

lib/src/services/offline_repository.dart

The fix adds await reconcileParentTableForMeta(doctype, tableName, parentMeta) before the write transaction in saveDocument. This mirrors what PullEngine already does before applying a pull page. The comment correctly notes it MUST run outside the transaction — ALTER TABLE in SQLite is DDL and auto-commits; running it inside a transaction would deadlock in WAL mode.

The causality is right: Fix 2 (cache invalidation) ensures parentMeta returned by _loadMeta is fresh; Fix 1 ensures the SQLite table schema matches that fresh meta before the INSERT is built. The two fixes compose cleanly — Fix 1 alone (with stale meta) would still reconcile against the old schema and skip the new column, so Fix 2 is necessary to make Fix 1 actually help. The PR describes this correctly.

One performance note: reconcileParentTableForMeta runs a PRAGMA table_info on every saveDocument call, even when no drift exists. PRAGMA table_info is O(columns) and very fast, so this is acceptable, and PullEngine already pays the same cost on every pull page. Not blocking, but worth knowing.

---
Fix 2 — invalidateMetaCacheFor wired to MetaService.onMetaRefreshed

lib/src/services/meta_service.dart, lib/src/services/offline_repository.dart, lib/src/sdk/frappe_sdk.dart

onMetaRefreshed fires from _upsertMetaJson — the single chokepoint through which every server meta rewrite passes (checkAndSyncDoctypes, resyncMobileConfiguration, prefetch*, getMeta's network branch). Placing the callback here is the right level — any higher and it would miss some callers; any lower and it would fire multiple times for the same doctype.

invalidateMetaCacheFor clears three caches keyed by doctype: _metaCache, _childMetasByParent, and _ensuredTables. The _ensuredTables removal forces the next pull to re-run ensureSchemaForClosure for the affected doctype, which is the correct belt-and-suspenders for the pull path (the save path already self-heals via Fix 1).

---
MEDIUM M1 — No end-to-end SDK wiring test

lib/src/sdk/frappe_sdk.dart line 14 (in the diff)

_metaService!.onMetaRefreshed = _repository!.invalidateMetaCacheFor is the critical glue that makes Fix 2 work in production. The existing tests cover both sides in isolation — meta_service_test.dart confirms onMetaRefreshed fires, and offline_repository_reconcile_test.dart confirms invalidateMetaCacheFor takes effect. But nothing tests that FrappeSDK actually sets this wiring up. If the assignment line in frappe_sdk.dart were accidentally removed or placed after the services are null, the integration would silently break with no test signal. A short test using FrappeSDK.forTesting that calls fetchAndStoreInDb and then asserts a subsequent saveDocument uses the fresh schema would close this gap.

---
MEDIUM M2 — dropdownFullTap hardcodes TextDirection.ltr for non-EdgeInsets contentPadding

lib/src/ui/widgets/fields/field_helpers.dart lines 186–189

final resolved = cp is EdgeInsets
    ? cp
    : cp?.resolve(TextDirection.ltr) ??
          const EdgeInsets.symmetric(horizontal: 12, vertical: 12);

When contentPadding is a DirectionalEdgeInsets or other EdgeInsetsGeometry subtype (which consumer-provided FieldStyle could supply), resolve(TextDirection.ltr) swaps start/end for left/right assuming LTR. In an RTL layout, left and right are swapped, so the padding redistribution would be visually incorrect — the left inset moves to the wrong side of the dropdown.

The function signature has no BuildContext, so the ambient TextDirection is not available. For the SDK's current deployment context (India, all LTR), this has no practical impact. But as a public utility function it should either accept a TextDirection parameter or document the LTR assumption explicitly in the doc comment. At minimum, add a note: "Assumes TextDirection.ltr when resolving directional insets — pass an explicit EdgeInsets if RTL support is needed."

---
LOW L1 — reconcileParentTableForMeta called with potentially null parentMeta guard

lib/src/services/offline_repository.dart

if (parentMeta != null) {
    await reconcileParentTableForMeta(doctype, tableName, parentMeta);
}

When parentMeta is null (meta was never synced for this doctype), reconcile is skipped and the function continues to the write transaction. At that point the INSERT will fail on the docs__ table anyway since it has no schema. This is a pre-existing condition — the save path already had a null-meta guard downstream. The behaviour is unchanged, just worth confirming the downstream guard still produces a clean error rather than a confusing crash.

---
LOW L2 — childMetasByParent invalidation misses parent when a child doctype is refreshed

lib/src/services/offline_repository.dart, invalidateMetaCacheFor

_childMetasByParent.remove(doctype) removes child meta keyed by this doctype as a parent. If onMetaRefreshed fires for a child doctype (e.g. "Order Item") because its meta changed server-side, _childMetasByParent['Order Item'] is cleared, but _childMetasByParent['Order'] (which holds the cached child meta for the "Order" parent) is not. If the "Order Item" schema changed (new child field), the parent's cached child meta entry is stale until the next _loadMeta re-read of the parent. This is a pre-existing design limitation, not introduced by this PR, but the new callback surfaces it as a reachable scenario for the first time.

---
Fix 3 — dropdownFullTap() — full-box tap target

The implementation is correct. Zeroing contentPadding and moving the inset into the dropdown's own padding (which is part of the clickable area) is the right mechanism. isExpanded: true is the correct addition — without it, the dropdown may not fill the available width and the padding redistribution would not cover the full box edge-to-edge.

The four widget tests (left edge, center, top edge, bottom edge) all tap within the old dead zones and assert the menu opens. The 8px from left and 4px from top/bottom offsets are well inside the old 16px/12px dead margins. The tests are correctly structured.

Omprakash-48:
<pre><font color="#00B4FF"><b>PR #80 — Round-1 review follow-up (</b></font><font color="#00FF88"><b>9041c29</b></font><b>)</b>                                                                                                             
                                                                                                                                                         
 <font color="#00B4FF"><b>### TL;DR</b></font>                                                                                                                                               
                                                                                                                                                         
 The three original fixes were already correct. This commit (a) hardens the one with a <b>real latent bug the review under-stated</b> (the in-txn               
 meta-resolution deadlock), and (b) closes the test/doc gaps. No behavioral change to the happy path.                                                    
                                                                                                                                                         
 <font color="#00B4FF"><b>### ✅ Fixed / addressed</b></font>                                                                                                                                
                                                                                                                                                         
 ┌─────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐ 
 │ <b>Review item            </b> │ <b>Disposition                                                                                                              </b> │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>L1</b> (null <font color="#00FF88">parentMeta</font>)    │ <b>Root-caused &amp; fixed.</b> Correction: this was <b>not</b> a &quot;clean INSERT failure&quot; — <font color="#00FF88">writeParentInTxn</font> resolved meta <i>inside</i> the write  │ 
 │                         │ txn (<font color="#00FF88">MetaService.getMeta</font> reads the outer DB), which <b>hangs forever</b> on sqflite&apos;s single non-reentrant lock (a Dart mutex    │ 
 │                         │ deadlock, never <font color="#00FF88">SQLITE_BUSY</font>). There were <b>two</b> vectors: parent (<font color="#00FF88">parentMeta ??=</font>) and a per-child resolver fallback. Fix:     │ 
 │                         │ <font color="#00FF88">writeParentInTxn</font> never resolves meta in-txn — <font color="#00FF88">parentMeta</font> is now <font color="#00FF88">required</font>, unsupplied children are skipped (loud log);     │ 
 │                         │ <font color="#00FF88">saveDocument</font> throws a clean <font color="#00FF88">StateError</font> before the txn and pre-resolves child metas.                                       │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>M1</b> (no end-to-end       │ <b>Fixed.</b> <font color="#00FF88">forTesting</font> didn&apos;t even replicate the production <font color="#00FF88">onMetaRefreshed → invalidateMetaCacheFor</font> wiring. Extracted a       │ 
 │ wiring test)            │ shared <font color="#00FF88">_wireMetaRefreshInvalidation()</font> called by both <font color="#00FF88">_doInitialize</font> and <font color="#00FF88">forTesting</font>, + an end-to-end test                   │ 
 │                         │ (<font color="#00FF88">frappe_sdk_meta_refresh_wiring_test.dart</font>) asserting a refresh evicts the offline cache.                                  │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>M2</b> (LTR hardcode)       │ <b>Doc-only</b> (correct for current LTR deployment). Documented <font color="#00FF88">dropdownFullTap</font>&apos;s <font color="#00FF88">TextDirection.ltr</font> assumption. No code change. │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>L2</b> (<font color="#00FF88">_childMetasByParent │ </font><font color="#00FF88"><b>Doc-only.</b></font><font color="#00FF88"> Confirmed pull-path-only; the save path rebuilds child metas via _loadMeta</font> (which <i>is</i> evicted), so saves are     │ 
 │ <font color="#00FF88">parent-miss)            │ unaffected. Documented on invalidateMetaCacheFor</font>.                                                                         │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>Fix 1 / Fix 2 / Fix 3</b>   │ Verified correct as written — <b>no code change</b>. (Note: the Fix-1 review rationale &quot;ALTER TABLE auto-commits / WAL deadlock&quot; │ 
 │                         │ is inaccurate — SQLite DDL is transactional; the real reason reconcile runs outside the txn is sqflite single-queue       │ 
 │                         │ reentrancy, which the existing code comment already states correctly.)                                                    │ 
 ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤ 
 │ <b>C</b> (extra, not in        │ Routed <font color="#00FF88">MetaService.ensureUpToDate</font>&apos;s schema write through <font color="#00FF88">_upsertMetaJson</font> so <font color="#00FF88">onMetaRefreshed</font> fires. <b>Latent</b> —               │ 
 │ review)                 │ <font color="#00FF88">ensureUpToDate</font> has no production caller today; defensive only.                                                            │ 
 └─────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘ 
                                                                                                                                                         
 <font color="#00B4FF"><b>### ❌ Not fixed (deliberately out of scope)</b></font>                                                                                                            
                                                                                                                                                         
 <font color="#00B4FF">- </font><b>Mid-sync child-row preservation.</b> When a child&apos;s meta <b>and</b> table are both still absent (closure-expansion-vs-save race), its rows are <b>skipped (with a   </b>
 <b>loud log)</b>, not preserved. Note this is <b>strictly better than before</b> (previously: hang). Fully preserving them needs a pre-txn meta <b>+ table</b> backfill — a  
 separate, larger change. Marked in code.                                                                                                                
 <font color="#00B4FF">- </font><font color="#00FF88"><b>doc_type_meta.dart:118</b></font> <font color="#00FF88">use_null_aware_elements</font> info-lint — pre-existing, from PR #80&apos;s 4th commit (link-option translation), untouched.               
 <font color="#00B4FF">- </font><b>The 4th commit</b> (<font color="#00FF88">23ad158</font>, link-option title translation) — outside the Round-1 3-fix scope; not re-reviewed here.                                      
                                                                                                                                                         
 <font color="#00B4FF"><b>### Verification</b></font>                                                                                                                                        
                                                                                                                                                         
 <font color="#00B4FF">- </font>New L1 regression tests use a resolver that <b>blocks on an outer-DB read</b>, so any reintroduced in-txn resolution hangs → 5s timeout → test fails         
 (genuine teeth, not a throwing stub).                                                                                                                   
 <font color="#00B4FF">- </font><font color="#00FF88">flutter analyze</font> clean on all changed files; <b>full SDK suite green (1571)</b>; consuming SNF app green (433). </pre>

deepak-dhwani (changes requested):
PR #80 Review — Round 2 (independent verification + first review of untouched commits)

Round-1 fixes — all independently re-verified as correct

---------

New findings (from commits round-1 didn't review)

MEDIUM — unguarded translate() hook can crash the link-option dropdown instead of degrading gracefully (lib/src/services/link_option_service.dart, translation commit). Every other pluggable hook in this file (LinkFilterBuilder, getLinkFilterBuilder) is wrapped in try/catch under an explicit "host hook failures must never propagate" contract. The new translate(label) call has no such guard, and its underlying TranslationService.translateDelegate is a public, externally-settable field. If a host wires a translator that throws, the exception propagates out of getLinkOptions/getLinkOptionsOffline — degrading to an empty dropdown rather than falling back to the untranslated English label (strictly worse than doing nothing). Fix: wrap translate(label) in try/catch with untranslated-label fallback, matching the file's own safeHook pattern.

MEDIUM — the dropdown fix's stated root cause may not match Flutter's actual behavior, though the fix is harmless. Per flutter/flutter's current dropdown.dart, the tap-opening GestureDetector wraps the entire InputDecorator with HitTestBehavior.opaque — contentPadding only affects visual layout of the child, not the hit-test region. If that's accurate for the pinned Flutter SDK version too, the "16px dead zone" this commit describes wouldn't exist, and the new edge-tap tests could pass identically against the pre-fix code (i.e., not actually testing a real regression). This doesn't contradict the developer's on-device repro (moto e5, Android 8) — Flutter's dropdown hit-testing has changed across SDK versions, so the bug may have been real on an older Flutter/flutter_form_builder combination and since fixed upstream, or specific to a widget composition not visible in the diff. Recommend: a quick sanity check — run the new edge-tap tests against a git-stash of the pre-fix field_helpers.dart/select_field.dart to confirm they actually fail without the fix. Not blocking; the change is visually equivalent and doesn't introduce risk either way.

LOW — LinkField's static-options dropdown has zero edge-tap test coverage. The PR description says the fix was "applied to SelectField and the LinkField static-options dropdown" — true in code (both call dropdownFullTap() identically), but only SelectField got the 4 new edge-tap tests. test/ui/widgets/fields/link_field_static_options_test.dart predates this commit and has no tapAt/edge assertions.

LOW — DocTypeMeta.translatedDoctype reuses isSubmittable's narrow boolean coercion (v == 1 || v == true) instead of the codebase's own canonical parseBool util, which the codebase's own doc-comment says every model should use for Frappe's inconsistent 0/1/"0"/"1" boolean encoding. Consistent with a pre-existing pattern (not a regression), but worth a follow-up ticket.

LOW — translation tests only cover the online getLinkOptions path, not getLinkOptionsOffline. Both currently share the same _rowsToEntities internals so there's no live bug, but a future divergence would regress the offline path with zero test signal.

-------

The two new MEDIUMs are worth a quick follow-up (the translate-hook guard is a 5-minute fix; the dropdown mechanism check is a "run once and confirm" sanity test)

Omprakash-48:
PR #80 — Round-2 review follow-up (8b9f84f)

### TL;DR

Four of the five round-2 findings were confirmed against the code and fixed. The one MEDIUM that would have changed behavior (the dropdown "root cause") was **empirically refuted** — I reverted the fix and reran the tests to prove the dead zone was real — so no code change there. All work is one commit on `om/bug_fix`.

### ✅ Fixed

| Finding | Sev | Disposition |
|---|---|---|
| Unguarded `translate()` in `_rowsToEntities` | MED | **Fixed.** Wrapped the host `translate()` hook in try/catch with an untranslated-English-label fallback, matching this file's own `resolveFilters`/`safeHook` "host hook failures never propagate" contract. Confirmed the vector is real: `TranslationService.translateDelegate` is a public, host-settable field, wired through `frappe_sdk.dart` — a throwing delegate previously propagated out of `getLinkOptions`/`getLinkOptionsOffline` and emptied the picker. |
| LinkField static-options has no edge-tap tests | LOW | **Fixed.** Added a `full-field tap target` group (left/center/top/bottom) to `link_field_static_options_test.dart`, mirroring SelectField's. |
| `translatedDoctype` narrow boolean coercion | LOW | **Fixed.** `translatedDoctype` and `isSubmittable` now use the canonical `parseBool` (already used elsewhere in the same model) instead of `v == 1 || v == true`, so Frappe's string `"0"`/`"1"` encodings resolve correctly. Fixed both, since they shared the identical latent bug. |
| Translation tests cover only the online path | LOW | **Fixed.** Added `getLinkOptionsOffline` translation tests (translated + untranslated) plus throwing-hook fallback tests on **both** the online and offline paths — these also lock in the guard above. |

### ❌ Not changed (verified non-issue)

| Finding | Sev | Disposition |
|---|---|---|
| Dropdown "root cause may not match Flutter; tests may lack teeth" | MED | **Refuted empirically.** Ran the requested sanity check: reverted `field_helpers.dart` + `select_field.dart` (and separately `link_field.dart`) to their pre-fix state (`92ed301^`), kept the current tests, and reran → **`+1 -3`** (left-margin, top-edge, bottom-edge taps all fail; only the center tap passes). The 16px/12px dead zone is real in the pinned `flutter_form_builder` `FormBuilderDropdown` (which exposes the `padding` param the fix redistributes into — not the raw `DropdownButtonFormField` the review reasoned about). The tests have genuine teeth; no change needed. |

### Verification

- `flutter analyze` clean on all changed files (also fixed a pre-existing `use_null_aware_elements` info in `doc_type_meta.dart` that the pre-commit lint surfaced once the file was staged).
- Full SDK suite green: **1579** (was 1571 + 8 new tests).
- Teeth check on the **new** LinkField edge-tap tests: `+1 -3` against pre-fix code, then restored — they genuinely fail without the fix.
- Pre-commit hook (flutter lint + dart format) passed; pushed to `om/bug_fix` @ `8b9f84f`.

### Files

- `lib/src/services/link_option_service.dart` — translate-hook guard
- `lib/src/models/doc_type_meta.dart` — `parseBool` for `translatedDoctype` + `isSubmittable`
- `test/services/link_option_service_translation_test.dart` — offline + throwing-hook coverage
- `test/ui/widgets/fields/link_field_static_options_test.dart` — full-box edge-tap coverage

deepak-dhwani (changes requested):
### New finding — translation commit (never previously reviewed)

**MEDIUM — unguarded `translate()` hook can crash the link-option dropdown instead of degrading gracefully** (`lib/src/services/link_option_service.dart`). Every other pluggable hook in this file (`LinkFilterBuilder`, `getLinkFilterBuilder`) is wrapped in try/catch under an explicit "host hook failures must never propagate" contract. The new `translate(label)` call has no such guard, and its underlying `TranslationService.translateDelegate` is a public, externally-settable field. If a host wires a translator that throws, the exception propagates out of `getLinkOptions`/`getLinkOptionsOffline` — degrading to an **empty dropdown** rather than falling back to the untranslated English label (strictly worse than doing nothing).

**Fix**: wrap `translate(label)` in try/catch with untranslated-label fallback, matching the file's own `safeHook` pattern.

---
Other LOW findings

- LinkField's static-options dropdown has zero edge-tap test coverage. The PR description says the fix was "applied to SelectField and the LinkField static-options dropdown" — true in code (both call dropdownFullTap() identically), but only SelectField got the 4 new edge-tap tests (test/ui/widgets/fields/select_field_test.dart). test/ui/widgets/fields/link_field_static_options_test.dart predates this commit and has no tapAt/edge assertions.
- DocTypeMeta.translatedDoctype reuses isSubmittable's narrow boolean coercion (v == 1 || v == true) instead of the codebase's own canonical parseBool util, which handles Frappe's 0/1/"0"/"1" boolean encoding inconsistencies. Pre-existing pattern, not a regression — worth a follow-up ticket.
- Translation tests only cover the online getLinkOptions path, not getLinkOptionsOffline. Both currently share the same _rowsToEntities internals so there's no live bug, but a future divergence would regress the offline path with zero test signal.

---

The one new MEDIUM (translate-hook guard) is a 5-minute fix and doesn't block merge, nothing here is a regression or a functional break.

Omprakash-48:
PR #80 — Round-3 review follow-up (all findings already resolved in `8b9f84f`)

### TL;DR

Every finding in this round is **already fixed and pushed** in commit `8b9f84f` on `om/bug_fix` (in sync with `origin/om/bug_fix` — 0 ahead / 0 behind). This review was cut against a pre-`8b9f84f` revision. Proofs below are quoted from the current file contents + observed test runs — nothing here is from memory.

Nothing to change. To see the fixes on the PR, pull the latest `om/bug_fix` head (`8b9f84f`).

---

### MEDIUM — unguarded `translate()` hook → ✅ already guarded

The requested fix ("wrap `translate(label)` in try/catch with untranslated-label fallback, matching the file's own `safeHook` pattern") is already in the code, verbatim.

`lib/src/services/link_option_service.dart:152-168`:

```dart
final translate = _translate;
if (translateLabels && translate != null) {
  // The translate hook runs host code (TranslationService.translateDelegate
  // is a public, host-settable field) and can throw. A thrown hook must
  // never propagate out of the picker — that would degrade the dropdown to
  // empty instead of just showing the untranslated English label. Same
  // "host hook failures never propagate" contract as [resolveFilters] /
  // [safeHook]; on failure we keep the untranslated label.
  try {
    label = translate(label);
  } catch (e, st) {
    debugPrint(
      'LinkOptionService._rowsToEntities: translate hook threw for '
      '"$label" ($doctype) — using untranslated label. $e\n$st',
    );
  }
}
```

This mirrors the two existing guards the review references:
- `resolveFilters` — `link_option_service.dart:308-316` (try/catch → fall back to meta)
- `safeHook` — `link_option_service.dart:337-353` (try/catch → return null)

**Test proof (observed):** throwing-hook tests on **both** entry points assert graceful degradation:

`test/services/link_option_service_translation_test.dart` — `translate: (_) => throw StateError('host translator blew up')`, then `expect(out, isNotEmpty)` + label falls back to English. Online path (`getLinkOptions`) and offline path (`getLinkOptionsOffline`) each have one.

```
$ flutter test test/services/link_option_service_translation_test.dart
00:00 +8: All tests passed!
```

**Why the test has teeth (Dart semantics, not assumption):** without the `try/catch`, the injected `StateError` propagates out of `getLinkOptions`/`getLinkOptionsOffline`, so `await svc.getLinkOptions(...)` throws and the test fails. The test can only pass because the guard catches it.

**Factual correction for the reviewer:** the pre-fix failure mode was an **empty dropdown** (uncaught exception propagating out of the option fetch), not a crash — which matches the review's own body ("degrading to an empty dropdown").

---

### LOW — LinkField static-options dropdown edge-tap coverage → ✅ added

The static-options `LinkField` applies the same `dropdownFullTap()` + `isExpanded: true` as `SelectField` (`lib/src/ui/widgets/fields/link_field.dart:77,92`), and now has matching edge-tap tests.

`test/ui/widgets/fields/link_field_static_options_test.dart` — `group('full-field tap target', …)` with 4 `tapAt` tests (left margin / center / top edge / bottom edge):

```
$ flutter test test/ui/widgets/fields/link_field_static_options_test.dart --plain-name "full-field tap target"
00:00 +1: full-field tap target tap on the left padding margin opens the menu
00:00 +2: full-field tap target tap on the center still opens the menu
00:00 +3: full-field tap target tap near the top edge opens the menu
00:00 +4: full-field tap target tap near the bottom edge opens the menu
00:00 +4: All tests passed!
```

**Teeth proof (empirical, run last session):** reverted `link_field.dart` + `field_helpers.dart` to pre-fix (`92ed301^`), kept the tests, reran → `+1 -3` (left/top/bottom fail, only center passes), then restored. The tests genuinely fail without the fix.

---

### LOW — `translatedDoctype` narrow boolean coercion → ✅ switched to `parseBool`

`lib/src/models/doc_type_meta.dart:38,46`:

```dart
bool get isSubmittable => parseBool(metaData?['is_submittable']);
...
bool get translatedDoctype => parseBool(metaData?['translated_doctype']);
```

Was `v == 1 || v == true`; now uses the canonical `parseBool` (`lib/src/utils/frappe_json_utils.dart:7`, already used elsewhere in the same file at lines 78 and 150), so Frappe's `0 / 1 / "0" / "1" / true / false` encodings all resolve. Fixed `isSubmittable` too, since it shared the identical latent bug.

---

### LOW — translation tests only covered the online path → ✅ offline path now covered

Both entry points delegate to the same `_rowsToEntities` (`link_option_service.dart:96-101` online, `386-391` offline). Added `getLinkOptionsOffline` translation tests (translated + untranslated) plus the throwing-hook fallback test on the offline path, so a future divergence can't silently regress it. (Included in the `+8` run above.)

---

### Status

- **Commit:** `8b9f84f` — `fix(link-options): guard translate hook + parseBool meta flags (PR #80 round-2)`
- **Branch:** `om/bug_fix`, pushed — `git rev-list --left-right --count HEAD...origin/om/bug_fix` → `0  0`
- **Pre-commit:** flutter-lint + dart-format hooks passed.
- **Full SDK suite:** `1579` passed.

### Files (all in `8b9f84f`)

- `lib/src/services/link_option_service.dart` — translate-hook guard
- `lib/src/models/doc_type_meta.dart` — `parseBool` for `translatedDoctype` + `isSubmittable`
- `test/services/link_option_service_translation_test.dart` — offline + throwing-hook coverage
- `test/ui/widgets/fields/link_field_static_options_test.dart` — full-box edge-tap coverage

deepak-dhwani (approved):
Approving with no outstanding findings

Issue #82, opened 2026-07-02
Form: computed / fetch_from field values don't cascade (Frappe set_value parity)


ankitDhwani:
## Summary

`FrappeFormBuilder` does not cascade **programmatically-set** field values. When a
client-script `onFieldChange` returns a patch (a computed value, or a value derived from a
`fetch_from`), that value is written into form state and pushed to the widget via
`patchValue` — but the target field's **own** change handler never runs, so *its*
dependents don't recompute.

This diverges from Frappe Desk, where `frm.set_value(fieldname, value)` fires that field's
change trigger (`frappe.ui.form.on(dt, { fieldname(frm) { ... } })`) even for a programmatic
set. Chained computed fields "just work" on Desk; on mobile the chain stops after one hop.

## Frappe out-of-the-box behaviour (what we're matching)

On Desk, form scripts routinely rely on set-value cascades, e.g.:

```js
frappe.ui.form.on('Procurement Transfer', {
  dispatch_order_id(frm) { /* sets commodity_code (fetch_from) */ },
  commodity_code(frm) { /* fetches + sets uom */ },      // fires when commodity_code is set programmatically
  transfer_quantity(frm) { frm.set_value('no_of_bags', ceil(qty*100/weight)) },
  bag_weight(frm) { frm.trigger('transfer_quantity') },
});
```

Picking `dispatch_order_id` sets `commodity_code`, which **triggers** the `commodity_code`
handler → `uom`. On mobile this second hop does not fire.

## Root cause

In `_FrappeFormBuilderState`, the change pipeline writes the patch into `_formData`
**before** calling `patchValue`:

```dart
_formData.addAll(patches);
_formKey.currentState?.patchValue(_normalizePatchValues(patches));
```

`patchValue` → `FormBuilderFieldState.didChange` → `widget.onChanged`, so the patched
field's `onChanged` *does* re-enter the pipeline — but by then `oldValue == value` (the value
is already in `_formData`), and the pipeline's `oldValue != value` guard no-ops it. So the
programmatic set never triggers the field's own `onFieldChange`.

## Reproduction

Doctype with Data fields `a`, `b`, `c` and:

```dart
onFieldChange: (name, value, data) {
  if (name == 'a' && value == 'hello') return {'b': 'B'};
  if (name == 'b') return {'c': 'C'};   // should fire when b is set programmatically
  return null;
}
```

[...]

PR #83, opened 2026-07-02
feat(ui): cascade programmatic field changes (Frappe set_value parity)

ankitDhwani:
# PR Description (Updated)

## Summary

This branch contains **one feature** and **two unrelated fixes** that rode along on the same branch. Disclosed here per review; they can be split on request.

---

### 1. Feature – computed-field cascade (Frappe `set_value` parity)

**Files:**  
`lib/src/ui/widgets/form_builder.dart`, `doc/COMPUTED_FIELD_CASCADE.md`, `test/widgets/form_builder_cascade_test.dart`

- `cascadeProgrammaticChanges` flag: an `onFieldChange` patch re‑fires the patched field's handler so dependents recompute (default off = legacy).
- Loop safety: value‑equality breaker + `_maxProgrammaticCascadeDepth` cap.
- Round‑2/3 fixes:
  - Defer the self‑referential self‑key widget patch one frame (StackOverflow fix).
  - Guard the deferred self‑key echo regardless of the cascade flag (infinite‑hang fix).
  - Numeric value‑equality so `10` and `"10.0"` don't cause a spurious re‑fire.

---

### 2. Unrelated fix – API `_server_messages` extraction (`4232fdb`)

**Files:**  
`lib/src/api/utils.dart`, `test/api/server_message_extraction_test.dart`

Normalizes `_server_messages` parsing across payload shapes.  
**No overlap** with the cascade feature.

---

### 3. Unrelated fix – depends‑on numeric‑string coercion (`aba6b8f`)

**Files:**  
`lib/src/utils/depends_on_evaluator.dart`, `test/depends_on_evaluator_test.dart`

Coerces numeric‑string operands in relational comparisons.  
**No overlap** with the cascade feature.

---

## Tests

- Full suite green (**1589 tests**)
- Analyzer clean

---

# PR Comment Replies

## MEDIUM (bundling)

> Correct – `4232fdb` (api `_server_messages`, `utils.dart`) and `aba6b8f` (depends‑on numeric coercion, `depends_on_evaluator.dart`) are unrelated to the cascade feature, each with its own tests and zero file overlap.  
> I've updated the PR description to disclose all three groups.  
> Confirmed via `merge-base..HEAD` that these are the only two extras – nothing else is bundled.  
> Happy to split them into their own PRs if you'd prefer clean history over the disclosure.

---

## LOW (`_normForCascade`)

> Fixed in `5202ae0`.  
> Value‑equality now compares numerically when both operands parse as numbers (so `10` == `"10.0"`), falling back to trimmed‑string equality otherwise (Link ids, Select text, null).  
> Regression test added under cascade‑on – verified failing pre‑fix (extra re‑fire) and passing post‑fix.


deepak-dhwani (changes requested):
## PR #83 — feat(ui): cascade programmatic field changes (Frappe set_value parity) — Round 1 Review

---

### MEDIUM M1 — Doc/CHANGELOG wording implies native `fetch_from` cascades too, but it doesn't

`_handleFetchFrom` (`lib/src/ui/widgets/form_builder.dart:589`, unchanged by this PR) applies its
linked-doc patches via its own `setState` + `patchValue(...)` call, entirely bypassing
`_programmaticEchoGuard` and `_scheduleProgrammaticCascade`. So a field populated by the SDK's
*native* `DocField.fetchFrom` metadata (not a consumer's `onFieldChange` handler) still won't
re-fire a dependent field's `onFieldChange`, even with `cascadeProgrammaticChanges: true`.

Checked this against issue #82's actual repro and "fetch_from" phrasing — the issue is scoped to
values a consumer's `onFieldChange` handler derives and returns as a patch (its example:
`dispatch_order_id(frm) { sets commodity_code via fetch_from }`), not the SDK's own `fetchFrom`
field property. So this isn't a broken promise or a regression — it's correctly out of scope for
#82. But `doc/COMPUTED_FIELD_CASCADE.md`'s wording ("a fetch_from result") reads ambiguously
enough that a future reader could assume native fetch_from chains are covered too.

**Fix**: add a one-line "Known limitation" note to `doc/COMPUTED_FIELD_CASCADE.md` clarifying that
native `DocField.fetchFrom` patches (applied in `_handleFetchFrom`) don't go through this cascade,
and/or file a fast-follow issue if that's desired behavior later.

---

### LOW L1 — Self-referential patch edge case untested

If a consumer's `onFieldChange` handler returns a patch for the *same* field that's changing
(e.g. `if (name == 'a') return {'a': 'normalized_a', 'b': ...}`), `prior[key]` for that self-key
is captured *after* the field's own new value was already written into `_formData`, not the
pre-user-edit value. Traced through, this doesn't cause an infinite loop (value-equality still
converges), but it's a slightly non-obvious case worth a one-line comment or a test if it's a
pattern you expect consumers to use.

---

M1 is a documentation clarity, not a functional defect, fine to address in this PR

ankitDhwani:
Round-1 findings addressed in two commits.

**M1 — doc scope** (`e5a2a06`): Verified — `_handleFetchFrom` applies its linked-doc patches through its own `setState` + `patchValue`, bypassing both the echo guard and the cascade scheduler, so native `DocField.fetchFrom` patches indeed never cascade. `doc/COMPUTED_FIELD_CASCADE.md` now has an explicit **"Known limitation: native `fetch_from` patches don't cascade"** section (cascade covers exactly handler-returned patches, matching #82's scope), the opening sentence no longer says "derived from a `fetch_from`", and the CHANGELOG entry carries the same scoping. If cascading native `fetchFrom` is wanted later, that can be a separate issue.

**L1 — self-referential patch** (`2b6855a`): Writing the suggested test found more than an untested edge — **it crashed**. A handler that *rewrites* its own field's value (`'hi'` → `'HI'`) recursed unboundedly through the text field's `didChange`/`TextEditingController` notification stack and died with a `StackOverflowError` (handler re-entered ~28k times before the stack blew).

Bisection results:

| cascade flag | self-only patch | self + sibling patch |
|---|---|---|
| OFF | unbounded recursion | unbounded recursion |
| ON | unbounded recursion | unbounded recursion |

So this is a **pre-existing defect on the flag-off path**, not introduced by this PR — the cascade's value-equality convergence you traced is correct, but the crash happens a layer below it, in the synchronous widget-patch echo, before the cascade's loop breakers can matter. (The existing depth-cap test never caught it because its self-patching field only ever changes programmatically, post-frame — never from inside a user-edit dispatch, which is the trigger.)

Fix: the self-key's **widget** patch is deferred one frame (post-frame callback), outside the field's own change dispatch; `formData` still updates immediately and non-self keys keep the synchronous path. The comment you asked for is at the `prior` snapshot, and regression tests pin both flag states: OFF → handler fires exactly once; ON → exactly twice (edit + one convergent re-fire, no depth-cap involvement). Also documented in the doc's "Self-referential patches are safe and converge" section and the CHANGELOG `Fixed` list.

Full suite: 1,577 tests green; `dart format` clean; `flutter analyze` — no new issues (the one pre-existing `example/lib/main.dart` import error is untouched and predates this branch).

deepak-dhwani (changes requested):
One new edge case worth flagging (not blocking)

The deferred self-key echo only raises _programmaticEchoGuard if (cascade). When cascadeProgrammaticChanges is off (the default) and the field is typed (Check/Date/Rating, where FieldNormalizer changes representation) and the handler unconditionally re-asserts its own field on every call (a non-convergent handler — the same "misuse" pattern the depth cap exists to catch), this path has no depth cap at all, because the depth cap only lives inside _scheduleProgrammaticCascade, which is cascade-gated. Pre-fix, that exact misuse pattern would have hit the same StackOverflow (arguably a "fail loud" outcome); post-fix, it would instead loop forever across postFrameCallbacks — an infinite hang rather than a crash. This requires a fairly specific combination (typed + self-referential + non-convergent handler + flag off) and doesn't affect the actual scenario the dev tested (an idempotent rewrite on a Data field), so it's a LOW, not a blocker — but worth a one-line doc note or a depth-cap check that isn't cascade-gated.

Both round-1 findings are fixed, the new crash fix is real and correctly reasoned. The one new LOW is a narrow, not something introduced fresh by this fix.

Omprakash-48:
# Fix: Unconditional Deferred Self‑Key Echo Guard

## Problem Summary

**Confirmed finding:**  
The deferred self‑key echo raised `_programmaticEchoGuard` **only** under `if (cascade)`. Meanwhile, the sole loop breaker – the depth cap in `_scheduleProgrammaticCascade` – is itself cascade‑gated.  
Consequently, with **`cascade = false`** and a typed field, the echo path had **no bound**, leading to infinite loops.

---

## Trigger Refinement

*(Verified by running both cases pre‑fix)*

The loop does **not** occur merely with a non‑convergent handler.  
It requires that the **handler's returned representation differs from what the widget echoes**.

### Concrete Example: `Check` field

| Handler return          | Widget echo        | Result |
|-------------------------|--------------------|--------|
| `{flag: 1}` (int)       | `1`                | `oldValue == value` → converges in **1 call** |
| `{flag: true}` (bool)   | `1` (since `_formData` holds `true`) | `true != 1` forever → **infinite loop** (measured **21+ calls** pre‑fix before the test’s `throw‑after‑N` guard tripped) |

---

## Pre‑Fix Behaviour

- **Synchronous execution** → `StackOverflowError`  
- **Deferred execution** → uncapped post‑frame hang (no stack overflow, but infinite scheduling)

---

## Fix Applied

> **Raise `_programmaticEchoGuard` unconditionally** for the deferred self‑key echo.

- This makes the deferred self‑echo a **pure state‑sync no‑op**, irrespective of the `cascade` flag.
- The **synchronous cross‑field (rest) echo** remains cascade‑gated *by design* – a bad handler there causes a `StackOverflowError` (fail‑loud), which is acceptable per your noted behaviour.

### Rationale

No finite cascade‑off use case relies on the self‑echo re‑running the pipeline.  
Therefore, nothing that worked before regresses with this change.

---

## Testing

- **Regression test added:** `form_builder_cascade_test.dart`  
  - Uses a cascade‑off `Check` field with a self‑referential `{flag: true}` handler.
  - The test throws after N calls, ensuring a would‑be hang fails fast instead of timing out on `pumpAndSettle`.
- **Result:** Fails pre‑fix, passes post‑fix.
- **Full suite:** 1588/1588 tests green; analyzer clean.

deepak-dhwani (dismissed):
MEDIUM — Two unrelated fixes are bundled into this branch, undisclosed in the PR description

lib/src/api/utils.dart (_extractServerMessage shape normalization, aba6b8f) and lib/src/utils/depends_on_evaluator.dart (numeric-string coercion, 4232fdbb) landed on this branch after the Round 2 review, so neither review has covered them until now. The PR description still says "this touches only form_builder.dart" and "no file overlap with #80" — no longer accurate.

Reviewed both on their own merits — both are correct, well-tested, no issues found.

Fix: update the PR description to disclose these two fixes, or split them into their own PRs (zero file/logical overlap with the cascade feature).

---

LOW — _normForCascade string-normalizes rather than numeric-normalizes

"10.0" vs "10" compares unequal in the cascade's value-equality check, causing a harmless extra self-terminating re-fire. Not a correctness issue given the depth cap backstop — just worth a note if it shows up in a real chain.

Omprakash-48:
## **8c55150** – `fix(ui): guard deferred self‑key echos regardless of cascade flag` *(LOW from round‑3)*

The deferred self‑key echo raised `_programmaticEchoGuard` only under `if (cascade)`, while the only loop breaker (`_maxProgrammaticCascade`) lives in the cascade‑gated `_scheduleProgrammaticCascade`. So with the flag **off** + a typed field, a self‑referential handler re‑fired every post‑frame with no cap – an uncapped infinite hang (pre‑fix a sync `StackOverflowError`).

- The self‑key echo guard is now **unconditional** (pure state‑sync no‑op).
- The synchronous cross‑field (`rest`) echo stays cascade‑gated (fail‑loud).

**Precise trigger** (verified by running both pre‑fix):
- Handler returning `{flag: 1}` (int) → converges in **1 call**.
- Handler returning `{flag: true}` (bool) → loops (widget echoes `1`, `_formData` holds `true`).

**Test:** Regression test added under cascade‑off with a `throw‑after‑N` guard so a hang fails fast.

---

## **5202ae0** – `fix(ui): numeric value‑equality in programmatic cascade` *(LOW `_normForCascade`)*

Cascade value‑equality compared trimmed strings, so `10` vs `"10.0"` counted as a change and caused an extra self‑terminating re‑fire.

Now compares **numerically** when both operands are numbers, else falls back to trimmed‑string equality (Link ids, Select text, null).

---

## **2b23b87** – `docs(ui): changelog for the two fixes above`

---

## MEDIUM (bundling)

Confirmed via `merge-base..HEAD` that exactly **two unrelated fixes** run on this branch – `4232fdb` (api `_server_messages`, `utils.dart`) and `aba6b8f` (depends‑on numeric coercion, `depends_on_evaluator.dart`), each with its own tests, zero overlap with the cascade feature.  
PR description updated to disclose all three groups. Can split into separate PRs on request.

---

## Verification

- Full suite green (**1589 tests**)
- Analyzer clean

PR #105, opened 2026-08-03
feat(ui): Desk-parity depends_on evaluation and programmatic field cascade

Omprakash-48:
## Summary

`om/bug_fix` → `develop`. Two independent threads, both about making the mobile form behave the way Frappe Desk does:

1. **Computed-field cascade** — `frm.set_value` parity for programmatic field changes, behind a default-off flag, plus three defect fixes found in review rounds 1–3 (PR #83, merged in here).
2. **`depends_on` evaluator** — replaced the string/regex pattern-matcher with a real JavaScript-subset parser and interpreter, so `depends_on` / `mandatory_depends_on` / `read_only_depends_on` agree with Desk.

Plus one unrelated API fix (`_server_messages` extraction).

---

### Feature — `FrappeFormBuilder.cascadeProgrammaticChanges` (default `false`)

When a value is set programmatically by `onFieldChange` (a computed value returned by a handler), the SDK previously did **not** re-fire that field's change pipeline, so dependents never recomputed. The pipeline writes the patch into form state *before* calling `patchValue`, so the re-entrant `onChanged` sees `oldValue == value` and self-guards.

With the flag on, the patched field re-fires its own pipeline — mirroring Desk, where `frm.set_value` fires the change trigger.

**Loop safety:** value-equality (a field re-fires only when its value actually changed → fixpoint) plus a depth cap. The synchronous `patchValue` echo is suppressed while cascading so each handler fires exactly once — no double-run for typed fields whose representation `FieldNormalizer` rewrites.

Default-off, so legacy behaviour is unchanged for every existing consumer.

### Fixes found in review

**`2b6855a` — StackOverflow on a self-referential patch.** A handler patching the *same* field that is dispatching its change (`'hi'` → `'HI'`) recursed unboundedly through the text field's `didChange` / `TextEditingController` notifications. Pre-existing defect, independent of the flag (reproduced on the flag-off path). The self-key widget patch is now applied in a post-frame callback, outside the field's own change dispatch; `_formData` still updates immediately and non-self keys keep their synchronous path. Tests pin both flag states: OFF fires the handler once, ON exactly twice (edit + one convergent re-fire).

> ⚠️ This deferral is load-bearing. Making it synchronous reintroduces the StackOverflow — there is a test that proves it. Please don't "simplify" it.

**`8c55150` — uncapped re-fire loop with the flag OFF.** The deferred self-key echo raised `_programmaticEchoGuard` only when `cascadeProgrammaticChanges` was on. With the flag off and a typed field (Check/Date/Rating), the handler patch lands in `_formData` in one representation while the widget's `onChanged` echoes another, so `changed` stays true and a self-referential handler re-fires every post-frame **with no depth cap** (the cap lives only in the cascade-gated path) — an uncapped hang. The guard is now raised unconditionally for the self-key echo, making it a pure state-sync no-op. The synchronous cross-field echo stays cascade-gated, so a bad handler there still fails loud. Regression test throws after N calls so the loop fails fast instead of hanging.

**`5202ae0` — spurious re-fire from representation-only changes.** `_scheduleProgrammaticCascade` compared patched values as trimmed strings, so `10` vs `"10.0"` counted as a real change and triggered a harmless but avoidable extra self-terminating re-fire. Now compares numerically when both operands parse as numbers, falling back to trimmed-string equality for Link ids, Select text, and null.

### Known limitation (documented in `doc/COMPUTED_FIELD_CASCADE.md`)

Native `DocField.fetch_from` patches do **not** cascade. `_handleFetchFrom` applies its patches through a separate path that bypasses the cascade. The flag is scoped to handler-returned patches only.

---

### The problem

Frappe Desk evaluates an `eval:` condition as actual JavaScript. `frappe/public/js/frappe/utils/utils.js` builds:

```js
new Function(...variable_names, `let out = ${code}; return out`)
```

and `layout.js::evaluate_depends_on_value` calls it with `{ doc, parent }` in scope.

`DependsOnEvaluator` instead matched a hardcoded list of string shapes — `' == '`, `' && '`, `'[..].includes(doc.x)'`. Anything outside those shapes fell through to a truthy fallback that looked up **the entire expression text** as a fieldname, found nothing, and returned false — silently hiding the field.

Measured against the Frappe source, 18 of 26 probed expressions diverged.

### What was broken

| Category | Example | Old result |
|---|---|---|
| Multiselect / Table MultiSelect | `(doc.present_season \|\| []).some(r => r.season == "Kharif")` | always `false` |
| `.length`, `.includes()` on a list | `doc.ms.length > 0` | always `false` |
| Compact operators | `doc.a=='X'&&doc.b=='Y'` | always `false` |
| Unary negation | `!doc.x` | always `false` |
| Ternary, parenthesised operands | `doc.a ? doc.b=='1' : doc.c=='1'` | always `false` |
| Frappe helpers | `in_list([...], doc.x)`, `cint`, `flt`, `cstr` | always `false` |
| Child-row parent access | `parent.channel=='Email'` | always `false` |
| `==` against a list | `doc.ms == 'A'` where `ms == ['A']` | `false` (JS: `true`) |

Note that "always false" for `depends_on` means **the field is hidden**, which then feeds clear-on-hide and the hidden-field strip in `_handleSubmit` — so the failure mode was not merely cosmetic.

### The fix — `lib/src/utils/js_expression.dart`

A tokenizer, precedence-climbing (Pratt) parser, and interpreter for the JS subset these conditions actually use. `DependsOnEvaluator`'s public API is unchanged.

**Supported:** literals (incl. array literals), `doc.`/`parent.` member and index access, `.length`, a whitelist of list/string methods (`includes`, `indexOf`, `some`, `every`, `filter`, `map`, `find`, `join`, `trim`, `toLowerCase`, `toUpperCase`, `startsWith`, `endsWith`, `split`), single-expression arrow functions, unary `!`/`-`/`+`, the full binary and ternary operator set, and the Frappe globals `in_list` / `cint` / `flt` / `cstr` / `Boolean`.

**Deliberately unsupported** (throws → per-property default): statements, assignment, block arrow bodies, template literals, `new`, regex literals.

Two implementation details that matter:

- **`&&` and `||` return operands, not booleans.** This is what makes the common `(doc.rows || [])` guard work — it must yield the array, not `true`.
- **Parsed ASTs are cached** by source text, mirroring Frappe's own `eval_function_cache`. The evaluator runs per field per change; re-tokenizing on every keystroke of a large form would be a real regression.

### Per-property defaults on failure

Previously a failed evaluation returned `true` uniformly. An unparseable `mandatory_depends_on` therefore made a field **permanently mandatory with no way to satisfy it**. Now the default is per property: `depends_on` → visible, `mandatory_depends_on` / `read_only_depends_on` → false. `evaluate2`'s `defaultWhenEmpty` doubles as `defaultOnError`.

### `parent` and Frappe std fields

- `parent` is threaded through `FormController` and `FrappeFormBuilder`. With no parent context it aliases `doc`, exactly as Desk does (`parent = this.frm ? this.frm.doc : this.doc`).
- `FormController._seedDefaults` now seeds the std fields that `depends_on` reads but `DocType.fields` does not carry — `docstatus`, `name`, `owner`, `doctype`, `idx`, `__islocal`, `__unsaved`. Without this, `eval:doc.docstatus == 0` hid fields in reactive mode that the legacy `_formData` path showed. They are excluded from `buildSubmitData`, so **the save payload is byte-identical** — a blanket `initialData` copy was rejected precisely because leaking `modified`/`creation` risks Frappe `TimestampMismatchError`.

### Three deliberate departures — each pinned by a test

1. **`[]` is truthy under `eval:`, falsy in the bare form.** In JS `[]` is truthy, so `eval:doc.ms` shows the field when nothing is selected. Only Frappe's non-`eval:` branch applies `$.isArray(v) ? !!v.length : !!v`. Asserting otherwise would introduce a *new* divergence.
2. **Relational and membership operators bridge number ↔ numeric-string.** `FieldNormalizer` stringifies Int/Float/Currency/Percent, while Desk holds real numbers. Strict JS would make `doc.qty > doc.max_qty` compare `"10" > "9"` lexicographically (false) and `[5,6].includes(doc.assessmnt_no)` compare `5` to `"5"` (false). Both hide fields Desk shows. **This is the one place the evaluator intentionally differs from real JS**, and it is a data-type problem in the SDK's own normalization, not a JS-semantics problem.
3. **An expression missing the `eval:` prefix stays a key lookup.** Desk's final branch is `doc[expression]`, so `depends_on: "doc.a == 'X'"` (no prefix) is always hidden there. Evaluating it would show a field Desk hides.

Also: unknown globals (`frappe`, `cur_frm`, `locals`) resolve to a falsy sentinel that equals nothing, rather than throwing into the "show it" default — matching a production Desk where `developer_mode` is off. `.length` on a non-string/list yields `undefined` (as in JS), and a trailing `;` is tolerated since Frappe wraps the condition in `let out = …;`.

---

## 3. Unrelated fix

**`4232fdb` — `_extractServerMessage` normalization.** It only handled the string-container + map-element shape and returned null for list-of-JSON-strings and list-of-maps, so callers surfaced "Unknown Error" instead of the real backend validation message. All shapes are now normalized, with coverage.

---

## Verification

Beyond the unit suite, the evaluator rewrite was validated by a **differential test against the previous implementation**, using Node as the ground-truth oracle:

- Extracted **all 699 unique** `depends_on` / `mandatory_depends_on` / `read_only_depends_on` expressions from **103 doctype JSONs** across the workspace (Frappe core + app doctypes).
- Generated 253 data combinations per expression from each field's real Select options, the string literals in the expression, and edge values → **176,847 evaluations**.
- Ran old (pre-change) and new implementations over every pair.
- Adjudicated every disagreement with `node` executing Frappe's exact wrapper, `new Function('doc','parent', 'let out = <expr>; return out')`, with `in_list`/`cint`/`flt`/`cstr` and a production-shaped `frappe.boot` in scope.

| Metric | Result |
|---|---|
| Divergences old vs new | 9,976 |
| **New matches Desk** | **9,720 / 9,723** |
| Old matches Desk | 3 |
| Neither | 0 |
| `referencedFields` diffs | **0** |

The 3 remaining are departure #2 above (the `Int`-normalization case). `referencedFields` being unchanged for all 699 expressions means **the reverse-dependency graph is identical to before** — no change to which fields recompute on which edits.

The differential caught four regressions the unit tests could not have, all fixed in this PR: numeric membership on a real `Int` field (`assessment_form`), `frappe.*` erroring into visible, trailing semicolons failing to parse, and `parent.` refs stealing the empty-set subscribe-all fallback from the dependency graph.

**Suite:** 1704 tests pass. `flutter analyze`: no issues.

### Coverage gap, stated explicitly

No expression in the 699-expression corpus uses `.some()` or a `Table MultiSelect` row shape, and the generated data used `List<String>`, never `List<Map>`. The differential therefore proves **no regression on existing expressions**; the new multiselect/`.some()` support is covered by targeted unit tests only.

### Known limitation

`eval:frappe.utils.html2text(doc.footer || '') != ''` returns `true` where Desk returns `false` for an empty footer — unknown globals equal nothing, so `!= ''` is true. It occurs once, in Frappe core `email_account`, which is not a mobile-fillable doctype. Resolving it would require three-valued logic for unknown globals, which was judged not worth the complexity.

---

## Files changed

| File | Change |
|---|---|
| `lib/src/utils/js_expression.dart` | **new** — tokenizer + parser + interpreter (+1165) |
| `lib/src/utils/depends_on_evaluator.dart` | rewired onto the parser; public API unchanged |
| `lib/src/ui/form/form_controller.dart` | cascade fixes; `parentData`; std-field seeding |
| `lib/src/ui/widgets/form_builder.dart` | cascade flag + echo guards; `parent` threading |
| `lib/src/api/utils.dart` | `_server_messages` normalization |
| `doc/COMPUTED_FIELD_CASCADE.md`, `CHANGELOG.md` | docs + known limitation |
| `test/depends_on_evaluator_test.dart` | +523 — multiselect, JS-subset, Frappe conditions, defaults |
| `test/ui/form/form_controller_depends_on_scope_test.dart` | **new** — `parent` + std-field scope, payload exclusion |
| `test/widgets/form_builder_cascade_test.dart` | cascade regression coverage, both flag states |

## Compatibility

- `cascadeProgrammaticChanges` defaults to `false` — no behaviour change unless opted in.
- `DependsOnEvaluator`'s public API is unchanged; `parentData` and `defaultOnError` are optional named parameters.
- `FormController` gains an optional `parentData` parameter.
- Save payloads are unchanged (std fields are eval-only).
- Dependency-graph edges are unchanged (`referencedFields` diff: 0 across all 699 real expressions).

## Suggested review focus

1. The three deliberate departures from real JS — these are judgement calls, not bugs, and each has a test that will fail loudly if someone "corrects" them.
2. `FormController._stdEvalFields` — the whitelist and its exclusion from `buildSubmitData`.
3. The post-frame deferral in the self-key patch — load-bearing, do not inline.
4. Whether `parent` threading is wanted in this PR at all; no app doctype currently uses `parent.` in a `depends_on` (all four occurrences are Frappe core desk doctypes), so it is inert today and could be split out if you'd prefer a tighter diff.


deepak-dhwani (changes requested):
Real net delta: the evaluator rewrite (js_expression.dart +1165, d417/+~150), two cascade fixes (8c55150 unconditional self-key echo
guard, 5202ae0 numeric cascade equality), parent threading, std-fidefaultOnError, and unrelated format/import churn in three files.

---
BLOCKER B1 — reactive submit payload silently drops fields gated on std fields

lib/src/ui/form/form_controller.dart:580-582 strips _stdEvalFields from complete, and then complete is reused as the doc scope for the visibility sweep at
:598-611 (containers) and :614-627 (own depends_on).

Trace for a field with depends_on: "eval:doc.docstatus == 0" on a

- _computeUiState evaluates against _rawValues, which now does conseedDefaults block) → visible: true. The field renders, the userfills it.
- buildSubmitData evaluates against complete, which does not contaurns _undefined → _looseEquals(_undefined, 0) → false → removeWherereturns !false → field removed from the payload.

The field is visible, editable, validated, and then silently discarded on save. Worse for a Section Break/Tab Break gated the same way: the container sweep
drops every field under it.

This is the direct consequence of fixing the UI path (_rawValues) Before this PR both were consistently wrong (field hidden, nothingto lose); now they disagree in the direction that loses user data. eval:doc.docstatus == 0 / doc.__islocal are extremely common in Frappe meta.

Fix — keep the payload clean but evaluate against the full doc:

// payload must not carry std fields; the eval scope must.
final evalScope = <String, dynamic>{
  ...complete,
  for (final k in _stdEvalFields)
    if (_rawValues.containsKey(k)) k: _rawValues[k],
};
then pass evalScope (not complete) to all three DependsOnEvaluator.evaluate calls in buildSubmitData, still removeWhere-ing on complete. Add a test that pins
it: docstatus: 0 + a field with depends_on: 'eval:doc.docstatus ==itData(). The existing test atform_controller_depends_on_scope_test.dart:104 uses a meta with no depends_on, so it can't catch this.

---
HIGH H1 — .length special-case shadows a field named length

lib/src/utils/js_expression.dart:656 checks key == 'length' before:663. Form data is a Map, so for any DocType with a field literallynamed length (dimension fields — length/width/height — are routine in Frappe apps), doc.length resolves to _undefined instead of the field's value.

eval:doc.length > 0 → NaN > 0 → false, permanently. The field is hidden, and via B1's sweep also stripped from the payload. No error, no log — the expression
parses fine.

Fix — put the own-property lookup first; the length fallback then  genuinely absent (which is what JS does for {}.length anyway):

if (obj is Map) {
  return obj.containsKey(key) ? obj[key] : _undefined;
}
if (key == 'length') {
  if (obj is List) return obj.length;
  if (obj is String) return obj.length;
  return _undefined;
}

---
HIGH H2 — two paths return true regardless of defaultOnError

lib/src/utils/depends_on_evaluator.dart:103 and :134:

if (expression == null || expression.isEmpty) return true;   // :1
...
if (expr.isEmpty) return true;                                // :

evaluate2 guards null/empty before delegating, so :103 only bites ltOnError: false) callers — but :134 is reachable through evaluate2: mandatory_depends_on: "eval:" (or "eval: ") is non-empty, reaches evaluate, strips the prefix, and returns true. That is a permanently mandatory field with no
way to satisfy it — verbatim the failure mode the per-property defsk would SyntaxError on let out = ; return out and error out, sotrue isn't the Desk answer either.

Fix — return defaultOnError; on both lines.

---
MEDIUM M1 — unbounded sdkLog on deterministic evaluation failur

depends_on_evaluator.dart:141-153 logs on every failed evaluation, expression. These expressions are static DocType meta and thefailure is deterministic, but _computeUiState runs per field per change. One unparseable depends_on on a 60-field form means a log line — and, in the generic
catch, a $st stack-trace interpolation — on every keystroke. Dedupatic final Set<String> _loggedFailures) so each bad expression isreported once per process.

MEDIUM M2 — the largest behavioural change in the PR has no CHANGELOG entry

CHANGELOG.md gains exactly two lines, both about the cascade. Nothing for the evaluator rewrite — despite the PR itself measuring 18/26 probed expressions
changing result and ~9,976 divergences over the real corpus. For ah fields are visible / mandatory / read-only, and which fields reach the save payload" changing semantics is the single most consumer-visible thing here. Also undocumented: DependsOnEvaluator.parentData / defaultOnError,
FormController.parentData, _stdEvalFields seeding, and the three dlease add an ### Added / ### Changed block, and state the departures where consumers will find them (not only in the PR description).

MEDIUM M3 — the new regression test can't fail on a partial regression

test/widgets/form_builder_cascade_test.dart:387 asserts lessThanOrEqualTo(5) while the guard at :363 throws at > 5. The documented contract for the cascade-OFF
path is "the handler fires exactly once" — as written, 2, 3, 4, ord is later re-gated on the flag in a way that leaks a couple ofextra frames, this test stays green. Same at :392: anyOf(true, 1) doesn't pin which representation lands in the doc. Make them equals(1) and pick the intended
representation. (The cascade-ON test's bHandlerCalls == 0 / == 1 at — match them.)

MEDIUM M4 — .some() / List<Map> is the newest path and has the

You state this yourself: zero expressions in the 699-expression coe MultiSelect row shape, and the generated data was List<String>only. So the differential — the thing that gives real confidence here — proves no regression, not correctness of the new capability. The unit tests cover the
happy path. Worth adding: .some() over a List<String> (arrow parammember on a String → throws → default-on-error, is that intended?);a row Map missing the key; doc.rows[0].season; .some() with a zero-arg arrow; .filter(...) chained into .some(...).

MEDIUM M5 — a host-supplied controller never receives parentData

form_builder.dart:484-490: widget.controller ?? FormController(..., parentData: widget.parentFormData). A host that owns its FormController and passes
parentFormData gets parent silently aliased to doc with no warningn the injected controller, or assert/log when both are supplied.

---
LOW

- L1 — _looseEquals (js_expression.dart:960-963) takes both operands to primitives unconditionally, so two objects compare by value: [] == [] → true, and
doc.ms_a == doc.ms_b on two multiselects can be true. JS compares se equality too, so both are false there. _strictEquals gets thisright (identical); loose should short-circuit the same way when both sides are List/Map.
- L2 — cint/flt (:691-698) are num.tryParse-based, but Frappe's arr group-separator stripping. cint('12abc') → 0 here, 12 in Desk;flt('1,200') → 0 here, 1200 in Desk. Narrow, but worth a comment at minimum since the file's contract is Desk fidelity.
- L3 — parseJsExpression (:1042-1055) doesn't cache failures, so ae-tokenizes and re-throws on every evaluation — precisely theper-keystroke cost _astCache exists to avoid. Cache a sentinel (or the exception) for the negative case; pairs naturally with M1.
- L4 — === against a Check field flips mid-session. initialData cas onChanged emits bool (FieldNormalizer line ~44). So eval:doc.flag=== 1 is true on load and false after the user toggles it. Same normalization class as your documented departure #2, but not bridged for ===. Either bridge
bool↔0/1 in _strictEquals or document it alongside the other depar
- L5 — the flag-off path is no longer strictly "byte-identical". form_builder.dart:945 dropped its cascade && gate, so any depth-0 change arriving while the
deferred self-key patchValue is in flight is now swallowed to a stlf field's echo. Only reachable when a handler patches its ownfield, so I think it's the right trade, but the doc/CHANGELOG wording ("byte-identical when off") overstates it.
- L6 — string-literal escapes (:167-171) write the escaped char vee letter n. Documented as intentional; fine, just noting it's silent rather than a parse error.
- L7 — unrelated churn: lib/src/api/utils.dart (format only), testctions_test.dart (4 unused imports removed),test/widgets/table_multi_select_placeholder_test.dart (format only), and the source: ternary reflow at form_builder.dart:1021. Looks like dart format under a
different version. Harmless, but it makes the diff read bigger tha

---
Verdict

The evaluator rewrite is the right change, wellntial is better evidence than most PRs of this size carry. But B1 silently drops user-entered data from the save payload for a very common Frappe meta pattern, and it's a direct consequence of the std-field seeding this PR
adds — that has to land with a test before merge. H1 and H2 are thspectively. M1–M3 I'd want in the same pass; the LOWs can befollow-ups.

Suggested split, if you want a tighter diff: B1/H1/H2 + the evaluator go in one PR; parent threading (inert today by your own account — all four parent.
occurrences are Frappe core desk doctypes) comes out into its own,properly.

---

deepak-dhwani (changes requested):
This is a strong PR — the depends_on rewrite is genuinely good work, and the differential harness (699 real expressions × 253 data combos, adjudicated by node running Frappe's exact wrapper) is the standard I'd want on every evaluator rewrite. No blockers. But the parent threading bolted on beside it doesn't work, and its tests can only pass.

3 HIGH

- H1 — parent is never aliased to doc (form_builder.dart:448). effectiveParentFormData is non-nullable, so depends_on_evaluator.dart:156's parentData == null ? doc : ... alias branch is unreachable from the widget. Result: in reactive mode doc is _controller.values (live) while parent is _formData — a snapshot frozen at initState, since only the legacy path writes it. Reactive section/tab visibility does route through this (_buildReactive → _buildTabContent → _shouldShowField → _evaluateDepends). The invariant is tested at depends_on_evaluator_test.dart:557, but only via paths where _parentData really is null.
- H2 — didUpdateWidget never re-threads parentFormData (:1724–1780). Assigned once in initState; a host passing a fresh parent map on rebuild keeps evaluating against the mount-time map.
- H3 — parentData setter calls notifyListeners() from inside initState (:541). That branch exists specifically for a host-supplied controller — i.e. one that may already have listeners. Marked plausible, not confirmed: it turns on host listener behaviour I can't observe here.

4 MEDIUM

- M1 — departure #2's numeric bridge covers relational/membership/loose-== but not + or ===. eval:doc.a + doc.b > 40 with a=5,b=0 gives '50' > 40 → true; Desk says false. -, *, /, % are fine.
- M2 — docstatus/__islocal guards test containsKey, so an explicit null reintroduces exactly the hidden-draft-field bug. Two-word fix.
- M3 — pop()'s mutation forces a per-evaluation copy of every list (twice, via H1) to protect a case that can't occur: the one real corpus expression is split('.').pop(). And the copy is one level deep, so doc.rows[0].tags.pop() still reaches live state.
- M4 — evaluate2's onError is removed and evaluate's renamed; DependsOnEvaluator is exported at frappe_mobile_sdk.dart:122. Filed under Added, with CHANGELOG.md:44 still describing fixing a parameter this PR deletes, and the PR body claiming "public API is unchanged."

2 LOW — the description claims lib/src/api/utils.dart (commit 4232fdb is already an ancestor of the base; zero diff); three comment/code drifts.

Two caveats on my end: no Dart/Flutter toolchain on this machine, so flutter test/flutter analyze were not run — the "1704 tests pass" claim is unverified by me, though no finding above depends on a test result. And the description's own offer to split the parent threading out is worth taking: it's inert in app doctypes today, so it buys nothing by riding along, and the evaluator could merge now.

Omprakash-48:
# PR 105 — complete list of changes made in response to review

Branch head `96ee234`. Pushed to `origin/om/bug_fix` and the fork; both in sync.

**Gates on this head:** `flutter test` **2131 passed** (exit 0) · `dart format
--set-exit-if-changed .` **460 files, 0 changed** · `flutter analyze` **1 error,
pre-existing and unrelated** (`example/lib/main.dart` imports
`config/app_config.dart`, gitignored at `.gitignore:75`; only
`app_config.example.dart` is checked in, so this reproduces on any fresh clone).

**Total diff vs base:** 14 files, +3949 / −828.

**How these were verified.** Every item below was checked by reading the actual
code, not by grepping for the fix. Findings marked *(executed)* were
additionally proved by running a throwaway probe — asserting the buggy behaviour
is gone *and* that the correct-negative still holds, so a fix can't pass by
being a blanket no-op. Probes were deleted after running; they are not in the
diff.

---

## Round 1 — all 15 findings addressed

| ID | Finding | Status | Where |
|----|---------|--------|-------|
| **B1** | Reactive submit payload silently dropped fields gated on std fields | **Fixed** *(executed)* | `5419b17` — `form_controller.dart:678` builds `evalScope`; all **three** `DependsOnEvaluator.evaluate` calls take it (`:705` tab, `:712` section, `:727` own-`depends_on`), `removeWhere` still on `complete` |
| **H1** | `.length` special-case shadowed a field named `length` | **Fixed** *(executed)* | `e3e29b6` — `js_expression.dart:672-675`, Map own-property lookup moved above the `length` case at `:676` |
| **H2** | Two paths returned `true` regardless of `defaultOnError` | **Fixed** *(executed)* | `e3e29b6` — `depends_on_evaluator.dart:110` and `:147` |
| **M1** | Unbounded `sdkLog` on deterministic evaluation failure | **Fixed** | `e3e29b6` — `_logOnce` + `_loggedFailures` (`:202-220`), bounded at 512, `resetLogGateForTest` seam |
| **M2** | Evaluator rewrite had no CHANGELOG entry | **Fixed** | `1ae690b`, `2b23b87` — CHANGELOG `:175-194` and `:247-275` |
| **M3** | Cascade regression test couldn't fail on a partial regression | **Fixed** | `04174a1` — `form_builder_cascade_test.dart:394` and `:403` now `equals(1)`; removed the `anyOf(true, 1)` / `lessThanOrEqualTo(5)` that `8c55150` had introduced |
| **M4** | `.some()` / `List` was the newest path with the least coverage | **Fixed** *(executed)* | `e3e29b6` — ~15 `.some()` tests: arrow params, missing key, empty list, zero-arg arrow, `(r,i)` index, `.filter().some()`, `doc.rows[0].season`, String receiver → default-on-error in both directions |
| **M5** | Host-supplied controller never received `parentData` | **Fixed** | `04174a1` — `form_builder.dart:541-543`, no-clobber (a controller already carrying `parentData` wins) |
| **L1** | `_looseEquals` compared two objects by value | **Fixed** *(executed)* | `e3e29b6` — `js_expression.dart:1001-1003`, `identical(a, b)` when both sides are List/Map |
| **L2** | `cint`/`flt` were `num.tryParse`-based | **Fixed** *(executed)* | `e3e29b6` — real `parseInt`/`parseFloat` prefix parsing; `cint('12abc')==12`, `flt('1,200')==1200` |
| **L3** | `parseJsExpression` didn't cache failures | **Fixed** | `e3e29b6` — `_astFailureCache` written at `:1204` **and read + rethrown at `:1191-1192`** |
| **L4** | `===` against a Check field flipped mid-session | **Fixed** *(executed)* | `e3e29b6` — `_strictEquals:967-968` bridges bool ↔ 0/1 |
| **L5** | "byte-identical when off" overstated the flag-off path | **Fixed** | `2b23b87` CHANGELOG `:265` states the departure explicitly; `doc/COMPUTED_FIELD_CASCADE.md:145-147` carves out the exception |
| **L6** | String-literal escapes write the escaped char | **No action** — reviewer confirmed intentional |
| **L7** | Unrelated format/import churn | **No action** — already in branch; format gate now passes clean |

Beyond the review, `5419b17` also fixed a second defect on the same seam: the
std fields were seeded only `if (initialData.containsKey(k))`, so a **new** local
doc carried no `docstatus` at all and a draft-only field was hidden before the
payload question even arose. Now `putIfAbsent('docstatus', () => 0)` plus
`__islocal = 1` when there is no name, matching `create_new.js:310`.

---

## Round 2 — 2 fixed, 3 confirmed and deferred, 1 open, 1 unverified

| ID | Finding | Status |
|----|---------|--------|
| **H1** | `parent` never aliased to `doc` in the widget path | **Confirmed** *(executed)* — deferred to the split |
| **H2** | `didUpdateWidget` never re-threads `parentFormData` | **Confirmed** — deferred to the split |
| **H3** | `parentData` setter notifies from `initState` | **Confirmed, upgraded from "plausible"** *(executed)* — deferred to the split |
| **M1** | Numeric bridge missing on `+` (and, additionally, `===`) | **Confirmed** *(executed)* — **open, not fixed** |
| **M2** | `containsKey` guard lets an explicit null reintroduce B1 | **Fixed** `843b86b` *(executed)* |
| **M3** | `pop()` forces a per-evaluation list copy | **Not verified** — no measurement taken, no claim made |
| **M4** | Breaking change to exported API filed under Added | **Fixed** `96ee234` |
| **LOW** | `lib/src/api/utils.dart` claimed in the description | **Confirmed** — zero diff vs base (see correction below) |
| **LOW** | Three comment/code drifts | **Not examined** |

### What the two fixes changed

**`843b86b` — M2.** `_seedDefaults` keyed on `initialData.containsKey(k)`, so an
explicit `{'docstatus': null}` stored the null and `putIfAbsent` then declined to
fire because the key was present. `null == 0` is false, so the field was hidden
*and* stripped from the save — the same data loss B1 fixed, reached through a
different door:

```
docstatus: null   -> visible=false  payload={draft_only: ''}     <- before
docstatus: null   -> visible=true   payload={draft_only: <value>} <- after
docstatus absent  -> visible=true                                 <- control
__islocal: null   -> now also correct
```

Seeding is on a non-null value now. `__islocal` inherits it (its derivation
guard reads `_rawValues`). `buildSubmitData`'s eval scope moved to the same
non-null test so a null cannot shadow `DependsOnEvaluator._evalScope`'s
`docstatus` default. Three regression tests added to
`form_controller_depends_on_scope_test.dart`.

**`96ee234` — M4.** `DependsOnEvaluator` is exported from the barrel
(`frappe_mobile_sdk.dart:122`) on **both** this branch and the base, so the
signature is public API:

```
origin/develop:  evaluate (…, {bool onError = true})
                 evaluate2(…, {bool onError = true})
here:            evaluate (…, {parentData, bool defaultOnError = true})
                 evaluate2(…, {parentData})            <- onError removed
```

Now filed under **Changed** as BREAKING with the migration (rename on
`evaluate`, drop on `evaluate2` — the third positional argument already carries
the intent) and the note that callers who never passed it are unaffected. Also
corrected the earlier entry that still described `onError` by its pre-rename
name.

### Evidence for the deferred three

H1, on a reactive Section Break gated `eval:parent.trigger == "GO"`:

```
parent.trigger gate : "under" field widgets  0 -> 0    <- Desk would show it
doc.trigger control : "under" field widgets  0 -> 1    <- finder works
controller path     : visible false -> true            <- alias fine here
```

The control matters — `0 -> 0` is a real miss, not a broken finder. The
controller path passing is exactly why the existing tests can't catch this.
Mechanism: `effectiveParentFormData` (`form_builder.dart:448`) is non-nullable,
so the evaluator's alias branch is unreachable from the widget; `_evalData`
(`:792-795`) returns live `_controller.values` while `parent` gets `_formData`,
and `_formData` is written only by `_onFieldValueChanged`, wired at `:1393` —
the legacy path only.

H3, with a host whose listener calls `setState` (the ordinary reason to own the
controller):

```
host listener fired during mount -> true
exception -> setState() or markNeedsBuild() called during build.
```

M1: `"5" + "0" > 40` is `true` here (Desk: false); `-` and `*` bridge correctly.
`===` has the same gap — `doc.a === 5` with `a="5"` is `false` while `==` is
`true`. So departure #2 is short on two operators, not one.

---

## Still open

1. **H1 / H2 / H3** — all parent threading. Proposed for the split, since H1's
   fix is a design choice (make `effectiveParentFormData` nullable so the alias
   fires, vs. make reactive `parent` track `_controller.values`) with different
   blast radii across 11 call sites. That decision shouldn't ride along with an
   evaluator change that is independently verified.
2. **Round-2 M1** — the `+` / `===` numeric bridge gap. Confirmed, not fixed.
3. **A gap found while writing the M2 tests, deliberately left** — recorded in
   `843b86b`'s message. `_computeUiState` reads `_rawValues` directly, so a null
   std field arriving *after* construction still mis-gates, and the controller
   then **clears the now-hidden field's value**. Reaching it needs
   `setValue('docstatus', null)` on a non-form field, so it is contrived, but it
   is real and belongs with the follow-up.
4. **The PR description** still says "public API is unchanged", now contradicted
   by the CHANGELOG. Needs a manual edit.

## One correction to the review

The `lib/src/api/utils.dart` LOW is right — `git diff origin/develop...HEAD --
lib/src/api/utils.dart` is empty, confirmed. But the stated reason isn't:
`4232fdb` exists locally, yet `git merge-base --is-ancestor 4232fdb
origin/develop` returns NO. Worth citing the empty diff rather than the
ancestry.

## Commits

| SHA | Subject |
|-----|---------|
| `e3e29b6` | fix(utils): match Desk on five depends_on value semantics |
| `5419b17` | fix(ui): stop dropping depends_on-gated fields from the save payload |
| `8c55150` | fix(ui): guard deferred self-key echo regardless of cascade flag |
| `5202ae0` | fix(ui): numeric value-equality in programmatic cascade |
| `04174a1` | fix(ui): wire parentData into a host-supplied controller; pin cascade tests |
| `1ae690b` | docs(changelog): record the depends_on evaluator rewrite and its departures |
| `2b23b87` | docs(ui): changelog for round-3 cascade fixes |
| `b952a1f` | fix(utils): actually port develop's fieldname-shape gate into extractEvalDocField |
| `843b86b` | fix(ui): seed std eval fields on a non-null value, not key presence |
| `96ee234` | docs(changelog): file the DependsOnEvaluator parameter change as breaking |

deepak-dhwani (changes requested):
Verified fixed

New — HIGH

H4 — the M2 fix landed in the opt-in engine only. 843b86b changes FormController, which is reactive mode. FormBuilderMode.legacy is the constructor default (form_builder.dart:280), documented as "the rollout safety valve" — and the legacy path never touches _seedDefaults. It seeds its own map unfiltered at :466, so {'docstatus': null} lands in _formData, reaches _evalScope — unchanged, still if (!src.containsKey('docstatus')) — and _looseEquals(null, 0) returns false. Field hidden. And the payload half follows the same route (_buildCompleteFormData merges _formData at :1865, skips the field at :1882), so it's the full original failure mode, not just the visible half.

The commit message reasons about this exact seam — "so a null cannot shadow DependsOnEvaluator._evalScope's docstatus default there" — and routes around _evalScope in one caller rather than fixing it. One line at the choke point (if (src['docstatus'] == null)) closes both engines, both legacy scopes, and the post-construction setValue gap you disclosed.

Oddity worth knowing: the second legacy sweep at :1921 is immune, because dataForDepends is rebuilt from meta.fields only and carries no docstatus at all. So the legacy engine disagrees with itself about whether the key exists.

Confirmed open, deferred

H1–H3 all reproduced by the author, H3 upgraded from plausible to confirmed. Deferring is right and the stated reason is the honest one. One condition: parentData/parentFormData ship in this PR, so the entry point goes live ahead of the working implementation — either hold the surface back or land a skip:-marked guard test.

Corrections

You're right about L1 and I was wrong. 4232fdb is not an ancestor of develop — the content arrived by another path, so the file is byte-identical while the SHA is branch-local. The empty diff is the correct citation. Finding stands, reasoning withdrawn.

On M3: fair that I made no measurement. The load-bearing half was correctness, not perf — the copy is one level deep, so doc.rows[0].tags.pop() still mutates live child-table state, and the one real corpus expression (split('.').pop(), your test at :503) never touches form state. Non-mutating pop removes both.

One line short. H4 is the only merge blocker and it's smaller than anything in round 1.

Omprakash-48:
You were right, and the one-line version is the right fix. H4 is closed at the
choke point you named. Head is `1c768a4`.

## H4 — fixed at the choke point

Reproduced before touching anything. The new test runs across
`FormBuilderMode.values`, and on the old code exactly one side failed:

```
FormBuilderMode.legacy  : draft_only in payload -> false   <- FAIL
FormBuilderMode.reactive: draft_only in payload -> true    <- passed (843b86b)
```

That is your finding precisely: the fix had landed in the opt-in engine and the
constructor default was still on the original failure mode, both halves of it.

The fix is the line you proposed, in `_evalScope`:

```diff
- if (!src.containsKey('docstatus')) 'docstatus': 0,
+ if (src['docstatus'] == null) 'docstatus': 0,
```

It closes more than legacy. Also now passing:

- both legacy save-path scopes;
- the post-construction `setValue('docstatus', null)` gap I disclosed as open in
  `843b86b` — `_computeUiState` hands its raw values straight to the evaluator,
  so the choke point catches that too. The test I had pulled for failing is
  back, and green.

Correct-negatives asserted in both engines (`docstatus: 1` still hides the
field), so the fix can't pass by showing everything.

**On your oddity:** confirmed, and worth writing down — `dataForDepends` is
rebuilt from `meta.fields`, so the second legacy sweep carries no `docstatus` at
all and was already immune. The legacy engine disagreed with itself about
whether the key existed. Noted in the commit message for whoever touches that
path next.

## M3 — taken, on your correctness argument

You're right that the load-bearing half was correctness, not perf. `pop()` is
non-mutating now and the one-level copy is gone: nothing in the grammar mutates
its receiver any more, so there is nothing left for the copy to protect. That
removes the nested hole (`doc.rows[0].tags.pop()` was deleting from live
child-table state on every keystroke, which the copy never reached) and the
per-evaluation allocation together. Added as documented departure #5.

One consequence worth flagging: the test that proved doc/parent aliasing did it
*through* pop's side effect (`doc.rows.pop() == 3 && parent.rows.length == 2`),
so it went red. It now asserts the aliasing directly — `doc.rows == parent.rows`,
which is reference equality on two lists after the `_looseEquals` fix — with the
negative alongside (two distinct equal-content lists compare false) so it can't
pass by `==` being vacuously true. Arguably a better test than the original: it
states the invariant instead of a side effect of it.

## H1–H3 — condition met

Landed the skip-marked guard tests
(`test/ui/form/form_builder_parent_threading_guard_test.dart`). All three encode
the behaviour Desk has, all three are reproduced-failing rather than
speculative, and each carries the mechanism in a comment. `skip:` comes off when
the follow-up lands; they should pass unchanged.

I went with guard tests over holding the surface back because the API is already
public in this release and pulling it is the larger change — but if you'd rather
the surface not ship at all ahead of the implementation, say so and I'll pull it
instead.

## Thanks for the L1 withdrawal

Appreciated, and your reading is the right one — content arrived by another path,
so the file is byte-identical while the SHA stays branch-local.

## Gates

2137 pass, 3 skipped (the guards above), `dart format` clean, `flutter analyze`
unchanged — one pre-existing `example/lib/main.dart` error from a gitignored
config file.

## Commits

| SHA | Subject |
|-----|---------|
| `7acf70d` | fix(utils): default docstatus at the evaluator choke point; stop mutating on pop |
| `1c768a4` | test(ui): guard the parent-threading surface that ships ahead of its fix |

## Still open, unchanged

- **H1–H3** — deferred to the split, now guarded.
- **Round-2 M1** — the numeric bridge is missing on `+` and `===`
  (`"5" + "0" > 40` is true here, Desk says false; `doc.a === 5` with `a="5"` is
  false). Confirmed, not fixed. Say the word if you want it in this PR rather
  than the follow-up.
- **The PR description** still claims "public API is unchanged", contradicted by
  the CHANGELOG since `96ee234`. Needs a manual edit.

deepak-dhwani (changes requested):
# Round 3 — `1c768a4` · Request changes

H4 and M3 are both fixed, at the choke point and in the shape discussed. The H1–H3 guard tests
satisfy the deferral condition. One finding blocks, and it is the structural sibling of H4 that I
missed when I raised it.

**Verification note:** static review only — no Dart toolchain on my side, so the reported test run
is not reproduced. `git diff 96ee234..1c768a4` touches `CHANGELOG.md`, `depends_on_evaluator.dart`,
`js_expression.dart` and four test files; `form_builder.dart` and `form_controller.dart` are
unchanged this round.

---

## 🔴 H5 — `__islocal` is H4 one field over: derived in reactive mode only

`docstatus` had **two** defences — `FormController._seedDefaults` *and* `_evalScope`'s default —
which is why legacy mode got it for free even before this round, and why moving the null test to
the choke point closed every path at once. `__islocal` has only the first.

```
$ grep -rn "__islocal" lib/
lib/src/ui/form/form_controller.dart      # ← and nothing else
```

`FormBuilderMode.legacy` is the constructor default (`form_builder.dart:280`) and never constructs
a `FormController`. Its `initState:466` is a bare `_formData.addAll(widget.initialData ?? {})`, and
`_evalScope` adds only `docstatus`:

```dart
// form_controller.dart:153 — reactive only
if (!_rawValues.containsKey('__islocal')) {
  final docName = _rawValues['name'];
  if (docName == null || docName.toString().isEmpty) _rawValues['__islocal'] = 1;
}

// legacy, brand-new document:
//   doc.__islocal        -> containsKey false -> _undefined
//   eval:!doc.__islocal  -> !truthy(undefined) -> TRUE
```

So on a new document in the default engine:

- `depends_on: "eval:!doc.__islocal"` — *"show once saved"* — evaluates **true** and **shows the
  saved-only field on create**. Desk holds `__islocal = 1` and hides it.
- `read_only_depends_on: "eval:!doc.__islocal"` — *"lock once saved"* — evaluates **true** and
  **locks a field the user is there to fill in.** This is the worse half, and it is the exact
  pairing your own comment at `form_controller.dart:146-151` gives as the reason the derivation
  exists.

Nothing tests the derivation outside `FormController`. The other two `__islocal` test sites both
pass the flag explicitly — `depends_on_evaluator_test.dart:346` with `{'__islocal': 1}`, and the
leak test with `'__islocal': 1` checking only that it stays out of the payload — so neither
exercises the derivation.

### Fix — same choke point, same shape

`name` is already in the scope, so the derivation transplants directly:

```dart
static Map<String, dynamic> _evalScope(Map<String, dynamic> src) => {
  ...src,
  if (src['docstatus'] == null) 'docstatus': 0,
  if (src['__islocal'] == null &&
      (src['name'] == null || src['name'].toString().isEmpty)) '__islocal': 1,
};
```

This makes `_seedDefaults`' derivation belt-and-braces rather than load-bearing, exactly as
happened with `docstatus`. It carries one assumption worth stating in the comment: a saved document
must have `name` in the eval scope or it reads as local. That is not a new assumption —
`_seedDefaults` already derives from the absence of `name` — so the two engines agree, which is the
point. Extend the `FormBuilderMode.values` loop you just wrote, with the correct-negative (a doc
*with* a name hides the new-only field) asserted in both engines.

### The one thing that would downgrade this to follow-up

If `eval:!doc.__islocal` occurs only in Frappe core desk doctypes the mobile form never renders,
it is inert today in the same way `parent.` is, and it belongs with H1–H3. You have the corpus from
the round-1 differential (103 doctype JSONs) — **that grep is the whole decision, and it is on your
side.** The CHANGELOG calls this form "the most common" and "the dominant corpus form" but does not
say whether the occurrences are core or app.

---

## 🟡 L3 — two maintenance guards left with the code they guarded

1. **The `...src`-first ordering is still load-bearing.** The old docstring carried *"The `...src`
   spread MUST come first … Reordering silently disables the copy."* The copy is gone; the hazard
   is not. Put `'docstatus': 0` before the spread and a `{'docstatus': null}` re-overwrites it —
   restoring H4 silently. As written it now reads like a conditional default that would naturally
   go first, which is the wrong instinct to leave unguarded.

2. **The read-only invariant needs its note where the edit happens.** `_evalScope`'s docstring says
   no method in the grammar writes to its receiver, which is now the only thing keeping the
   evaluator read-only. Someone adding `sort` / `reverse` / `push` / `splice` — all mutating in JS —
   edits `_listMethods` (`js_expression.dart:548`), not `_evalScope`. One line at the set.

---

## 🟡 M1 — carried, but leave it a written record

Correctly deferred, no objection. Departure #2 in the CHANGELOG gained a fifth sibling this round
and still reads as though the numeric bridge applies to every operator. One clause naming `+` and
`===` as the two it does not cover means the follow-up inherits documentation rather than a review
thread.

---

## 🟡 The PR description is byte-identical across all three rounds

Four claims are now known-wrong, two of them already acknowledged as needing a manual edit. The
description is what the merge commit and release notes are read from:

| Claim | Reality |
|---|---|
| "`DependsOnEvaluator`'s public API is unchanged" | Contradicted by the BREAKING entry in `96ee234` |
| §3 + files table list `lib/src/api/utils.dart` | Empty diff against base |
| "Three deliberate departures" / four in the table | CHANGELOG now lists five |
| No mention of the null-`docstatus` or `pop` fixes | Both landed in `7acf70d` |

---

## ✅ Verified fixed this round

**H4 — `docstatus` default moved to `_evalScope`.** Checked the three things that could silently
undo it: the conditional entry sits *after* the `...src` spread so `0` overwrites the spread's
`null` rather than the reverse; `0 == null` is false so a real draft keeps its `0`; the scope map is
still freshly allocated, so no `docstatus` key is injected back into `_formData` / `_rawValues`. The
test is the right shape — parameterised over `FormBuilderMode.values` with the correct-negative
(`docstatus: 1` still hides the field) asserted in **both** engines, so it cannot pass by
blanket-showing. Failing-on-legacy / passing-on-reactive against the old code is the discriminating
result.

**M3 — `pop()` non-mutating, copy removed.** Re-audited every List branch of `_method` to confirm
the copy is safe to drop: `includes` / `indexOf` / `join` / `some` / `every` / `find` read;
`filter` / `map` allocate; `JsArrayLiteral` allocates; `_toPrimitiveStatic` joins to a string;
`_member` indexes. Nothing writes to a receiver, so live child-table lists can be shared — including
the nested `doc.rows[0].tags` case the one-level copy never reached. The forced test rewrite is the
better test: it states the aliasing invariant (`doc.rows == parent.rows`, reference equality after
the `_looseEquals` object fix) instead of a side effect of it, with the negative alongside so it
cannot pass vacuously.

**H1–H3 deferral condition — met.** Three `skip: true` guard tests, one per finding, each
documenting the mechanism and the reproduction, with a file-level note that they are known-failing
rather than speculative. They convert to real coverage on the follow-up with no rewrite. This is
what I asked for; the `parentData` surface no longer ships unattended.

---

## Gate

| Area | R1 | R2 | R3 |
|---|---|---|---|
| JS subset parser | ✅ | ✅ | ✅ |
| Evaluator is read-only | 🟡 M3 | 🟡 M3 | ✅ Fixed |
| JS value semantics | 🟡 M1 | 🟡 M1 | 🟡 Deferred |
| `docstatus` seeding | 🟡 M2 | 🔴 H4 | ✅ Fixed, both engines |
| `__islocal` seeding | — | — | 🔴 **H5** |
| Payload integrity | ✅ | 🔴 H4 | ✅ |
| `parent` threading | 🔴 H1,H2 | 🔴 Confirmed | ✅ Deferred · guarded |
| Lifecycle / rebuild | 🟡 H3? | 🔴 H3 | ✅ Deferred · guarded |
| Public API surface | 🟡 M4 | ✅ Fixed | ✅ |
| Cascade loop safety | ✅ | ✅ | ✅ |
| PR description | 🟡 L1 | 🟡 Stale | 🟡 Stale |

**Requesting changes on H5 only** — and it resolves either way with one grep: if
`eval:!doc.__islocal` reaches an app doctype the mobile form renders, fix it here; if every
occurrence is core desk meta, say so and I'll approve with it moved to the follow-up alongside
H1–H3, M1, L2 and L3.

Worth naming the pattern, because it is the same defect twice: **H4 and H5 are both a std-field
default placed in one engine's seeding rather than at the evaluator both engines pass through.** H4
surfaced because the M2 fix went to the wrong layer; H5 was there the whole time and I missed it in
H4 by checking one field instead of the set. The general rule for the follow-up: anything
`_stdEvalFields` **derives** rather than merely passes through belongs in `_evalScope`. Today that
is exactly `docstatus` and `__islocal` — the other five are pass-through and need nothing.

Omprakash-48:
# Review response — all four rounds

Head `c0319ec`. Diff vs base: 15 files, +4386 / −828.

**Gates:** `flutter test` **2143 passed, 3 skipped** (the H1–H3 guards below) ·
`dart format --set-exit-if-changed .` **461 files, 0 changed** · `flutter
analyze` **1 error, pre-existing and unrelated** — `example/lib/main.dart`
imports `config/app_config.dart`, gitignored at `.gitignore:75`, with only
`app_config.example.dart` checked in, so it reproduces on any fresh clone.

**How these were verified.** Each finding was checked against the code rather
than grepped for a fix. Items marked *(executed)* were additionally reproduced
by a probe that asserts the buggy behaviour is gone **and** that the
correct-negative still holds, so a fix cannot pass by being a blanket no-op.
For the two engine-split defects (H4, H5) the discriminating result is recorded:
failing on `legacy`, passing on `reactive`, against the pre-fix code. Probes
were deleted after running and are not in the diff.

**Reviewer had no Dart toolchain** for rounds 2–4, so the runs below are not
reproduced on their side. Nothing in rounds 2–4 depended on a test result to be
established; the reproductions are stated so they can be re-run.

---

## Round 1 — 15 findings, all addressed

| ID | Finding | Status | Where |
|----|---------|--------|-------|
| **B1** | Submit payload silently dropped fields gated on std fields | **Fixed** *(executed)* | `5419b17` — `form_controller.dart:678` builds `evalScope`; all **three** evaluate calls take it (`:705` tab, `:712` section, `:727` own-`depends_on`); `removeWhere` still on `complete` |
| **H1** | `.length` special-case shadowed a field named `length` | **Fixed** *(executed)* | `e3e29b6` — Map own-property lookup moved above the `length` case |
| **H2** | Two paths returned `true` regardless of `defaultOnError` | **Fixed** *(executed)* | `e3e29b6` — `depends_on_evaluator.dart:110`, `:147` |
| **M1** | Unbounded `sdkLog` on deterministic failure | **Fixed** | `e3e29b6` — `_logOnce` + `_loggedFailures`, bounded at 512 |
| **M2** | Evaluator rewrite had no CHANGELOG entry | **Fixed** | `1ae690b`, `2b23b87` |
| **M3** | Cascade test couldn't fail on a partial regression | **Fixed** | `04174a1` — `equals(1)`, replacing the `anyOf(true, 1)` / `lessThanOrEqualTo(5)` |
| **M4** | `.some()` / `List` — newest path, least coverage | **Fixed** *(executed)* | `e3e29b6` — ~15 tests incl. missing key, empty list, zero-arg arrow, `(r,i)` index, `.filter().some()`, String receiver → default-on-error both directions |
| **M5** | Host-supplied controller never received `parentData` | **Fixed** | `04174a1` — `form_builder.dart:541-543`, no-clobber |
| **L1** | `_looseEquals` compared two objects by value | **Fixed** *(executed)* | `e3e29b6` — `identical(a, b)` when both sides are List/Map |
| **L2** | `cint`/`flt` were `num.tryParse`-based | **Fixed** *(executed)* | `e3e29b6` — real prefix parsing; `cint('12abc')==12`, `flt('1,200')==1200` |
| **L3** | `parseJsExpression` didn't cache failures | **Fixed** | `e3e29b6` — `_astFailureCache`, written **and** read + rethrown |
| **L4** | `===` against a Check field flipped mid-session | **Fixed** *(executed)* | `e3e29b6` — bool ↔ 0/1 bridge |
| **L5** | "byte-identical when off" overstated | **Fixed** | `2b23b87` CHANGELOG + `doc/COMPUTED_FIELD_CASCADE.md:145-147` carve-out |
| **L6** | String-literal escapes | **No action** — confirmed intentional |
| **L7** | Format/import churn | **No action** — format gate passes clean |

`5419b17` also fixed a second defect on the same seam that the review didn't
reach: std fields were seeded only on `containsKey`, so a **new** local doc
carried no `docstatus` at all and draft-only fields were hidden before the
payload question arose.

---

## Round 2 — 2 fixed, 3 confirmed and deferred, 1 open, 1 unverified

| ID | Finding | Status |
|----|---------|--------|
| **H1** | `parent` never aliased to `doc` in the widget path | **Confirmed** *(executed)* — deferred, guarded |
| **H2** | `didUpdateWidget` never re-threads `parentFormData` | **Confirmed** — deferred, guarded |
| **H3** | `parentData` setter notifies from `initState` | **Confirmed, upgraded from "plausible"** *(executed)* — deferred, guarded |
| **M1** | Numeric bridge missing on `+` — and also `===` | **Confirmed** *(executed)* — **open**, now documented |
| **M2** | `containsKey` lets an explicit null reintroduce B1 | **Fixed** `843b86b`, superseded by `7acf70d` |
| **M3** | `pop()` forces a per-evaluation list copy | **Fixed** `7acf70d` |
| **M4** | Breaking API change filed under Added | **Fixed** `96ee234` |
| **LOW** | `lib/src/api/utils.dart` claimed in the description | **Confirmed** — zero diff vs base |

**M4 detail.** `DependsOnEvaluator` is exported at `frappe_mobile_sdk.dart:122`
on **both** sides, so this is public API:

```
origin/develop:  evaluate (…, {bool onError = true})
                 evaluate2(…, {bool onError = true})
here:            evaluate (…, {parentData, bool defaultOnError = true})
                 evaluate2(…, {parentData})            <- onError removed
```

Now under **Changed** as BREAKING with the migration, and the older entry that
described `onError` by its pre-rename name is corrected.

**H1 evidence** — reactive Section Break gated `eval:parent.trigger == "GO"`:

```
parent.trigger gate : "under" field widgets  0 -> 0    <- Desk would show it
doc.trigger control : "under" field widgets  0 -> 1    <- finder works
controller path     : visible false -> true            <- alias fine here
```

The control matters: `0 -> 0` is a real miss, not a broken finder. The
controller path passing is why the existing tests couldn't catch it.

**H3 evidence** — host whose listener calls `setState`, the ordinary reason to
own the controller:

```
host listener fired during mount -> true
exception -> setState() or markNeedsBuild() called during build.
```

---

## Round 3 — H4, M3, and the deferral condition

**H4 — the M2 fix had landed in the opt-in engine only.** Correct, and the
one-line version at the choke point was the right fix. Reproduced across
`FormBuilderMode.values`; exactly one side failed:

```
FormBuilderMode.legacy  : draft_only in payload -> false   <- FAIL
FormBuilderMode.reactive: draft_only in payload -> true    <- passed (843b86b)
```

```diff
- if (!src.containsKey('docstatus')) 'docstatus': 0,
+ if (src['docstatus'] == null) 'docstatus': 0,
```

Closes legacy, both legacy save-path scopes, and the post-construction
`setValue('docstatus', null)` gap disclosed as open in `843b86b` —
`_computeUiState` hands raw values straight to the evaluator. That test is back
and green. Correct-negatives asserted in both engines.

*Oddity confirmed:* `dataForDepends` is rebuilt from `meta.fields`, so the
second legacy sweep carries no `docstatus` and was already immune — the legacy
engine disagreed with itself about whether the key existed.

**M3 — taken on the correctness argument.** `pop()` is non-mutating; the
one-level copy is gone. That closes the nested hole (`doc.rows[0].tags.pop()`
was deleting from live child-table state every keystroke, which the copy never
reached) and drops the per-evaluation allocation. Documented as departure #5.

One consequence: the test proving doc/parent aliasing did it *through* pop's
side effect, so it went red. It now asserts the invariant directly —
`doc.rows == parent.rows`, reference equality after the `_looseEquals` fix —
with the negative alongside so it can't pass vacuously.

**H1–H3 deferral condition — met.** Three `skip: true` guards in
`test/ui/form/form_builder_parent_threading_guard_test.dart`, one per finding,
each carrying its mechanism and reproduction, with a file-level note that they
are known-failing rather than speculative. `skip:` comes off when the follow-up
lands.

---

## Round 4 — H5, L3, M1 docs

**H5 — fixed here rather than deferred, because the grep came back split.**

You were right that the grep is the whole decision. It has two answers.

*chetna* — every occurrence is Frappe core desk meta:

```
frappe (core)   14 files — Letter Head, Web Page, DocType, User,
                           Data Import, Property Setter, Google Calendar…
mform_chetna     0
mobile_control   0
```

On that corpus your deferral offer was right and I'd have taken it.

*inops* — app meta, not core:

```
mgrant        8 × depends_on: "eval:!doc.__islocal"
                 periodic_impact/outcome/input/output/lfa_achievement,
                 quarterly_utilisation_report, rfp_invitation
sva_frappe    3 ×, including
              mandatory_depends_on:
                "eval:(doc.__islocal || (!doc.__islocal && doc.update_password))"
frappe_theme  2 ×
```

`periodic_*_achievement` and `quarterly_utilisation_report` are field-collection
doctypes — what this SDK renders. And `sva_user` fails the *opposite* way from
the read-only case you flagged: on a new doc that expression is false, so a
password field Desk makes **mandatory on create** was not.

Fix is your shape, in `_evalScope`:

```dart
if (src['__islocal'] == null &&
    (src['name'] == null || src['name'].toString().isEmpty))
  '__islocal': 1,
```

Reproduced first; exactly one side failed:

```
legacy  : new doc (no name) hides saved_only -> false   <- FAIL
reactive: new doc (no name) hides saved_only -> true    <- passed
```

Both correct-negatives asserted in both engines: a doc **with** a name still
shows the field, and an explicit host `__islocal` still wins (the derivation
fires only on a null). The `name` assumption is written into the docstring —
not new, since `_seedDefaults` already derived from the absence of a name, but
now stated.

**Your pattern is in the CHANGELOG as the rule:** a std field the SDK *derives*
belongs in the evaluator, where every engine passes through; one it merely
*passes through* does not. Today that is exactly `docstatus` and `__islocal`;
the other five need nothing.

**L3 — both guards landed.**

- `_evalScope`'s docstring reinstates the `...src`-MUST-be-first note. The
  hazard outlived the copy, and your read of the instinct is right: these now
  look like conditional defaults, which is what someone hoists to the top —
  and hoisting `'docstatus': 0` above the spread lets `{'docstatus': null}`
  overwrite it again, silently restoring H4.
- `_listMethods` carries the read-only invariant, at the set rather than in
  `_evalScope`, since that's where the edit happens. Names `sort` / `reverse` /
  `push` / `shift` / `unshift` / `splice` as mutating in JS, and states that
  the set is the only thing holding the property now the copy is gone.

**M1 — documented, still deferred.** Departure #2 now names the two operators
the bridge does **not** cover, with the divergence for each: `+` concatenates
(`"5" + "0" > 40` → true, Desk false); `===` compares types strictly
(`doc.a === 5` with `a="5"` → false, Desk true). Recorded as known gaps rather
than decisions.

---

## Corrections and acknowledgements

- **L1 (round 3), your withdrawal** — appreciated, and your reading is right:
  the content arrived by another path, so the file is byte-identical while
  `4232fdb` stays branch-local. The empty diff is the correct citation.
- **My round-2 note on that same LOW** — I said the finding stood but the
  ancestry reasoning didn't. That's the same conclusion, reached from the
  `merge-base` check.
- **H4 and H5 are the same defect twice**, as you named it: a derived std field
  defaulted in one engine's seeding instead of at the evaluator both engines
  pass through. H4 surfaced because my M2 fix went to the wrong layer; H5 was
  there the whole time. The CHANGELOG now carries the general rule so the
  follow-up doesn't repeat it a third time.

---

## Commits

| SHA | Subject |
|-----|---------|
| `e3e29b6` | fix(utils): match Desk on five depends_on value semantics |
| `5419b17` | fix(ui): stop dropping depends_on-gated fields from the save payload |
| `8c55150` | fix(ui): guard deferred self-key echo regardless of cascade flag |
| `5202ae0` | fix(ui): numeric value-equality in programmatic cascade |
| `04174a1` | fix(ui): wire parentData into a host-supplied controller; pin cascade tests |
| `1ae690b` | docs(changelog): record the depends_on evaluator rewrite and its departures |
| `2b23b87` | docs(ui): changelog for round-3 cascade fixes |
| `b952a1f` | fix(utils): actually port develop's fieldname-shape gate into extractEvalDocField |
| `843b86b` | fix(ui): seed std eval fields on a non-null value, not key presence |
| `96ee234` | docs(changelog): file the DependsOnEvaluator parameter change as breaking |
| `7acf70d` | fix(utils): default docstatus at the evaluator choke point; stop mutating on pop |
| `1c768a4` | test(ui): guard the parent-threading surface that ships ahead of its fix |
| `c0319ec` | fix(utils): derive __islocal at the evaluator choke point too |

---

## Open — carried to the follow-up

1. **H1 / H2 / H3** — parent threading. Deferred by agreement, guarded by
   skip-marked tests. H1's fix is a design choice (nullable
   `effectiveParentFormData` so the alias fires, vs. reactive `parent` tracking
   `_controller.values`) with different blast radii across 11 call sites.
2. **Round-2 M1** — the `+` / `===` numeric bridge. Confirmed, documented, not
   fixed.
3. **PR description** — stale on four counts, being corrected by hand:
   - "`DependsOnEvaluator`'s public API is unchanged" → contradicted by the
     BREAKING entry in `96ee234`
   - §3 and the files table list `lib/src/api/utils.dart` → empty diff vs base
   - "Three deliberate departures" / four in the table → the CHANGELOG now
     lists five
   - No mention of the null-`docstatus`, `__islocal` or `pop` fixes → landed in
     `7acf70d` and `c0319ec`

deepak-dhwani (approved):

PR #107, opened 2026-08-17
feat(attachments): offline attachment pipeline with on-device media cache

Omprakash-48:
Attachments picked on a device that is offline (or in offline-first mode) are now staged locally, queued, uploaded by the sync pass, and relinked to the real record and field on the server — with the uploaded bytes kept on device as the preview cache, so an attachment stays viewable without connectivity. Previously an attach field decided by connectivity alone: an online device uploaded inline, which put the file outside the offline pipeline entirely, and an offline pick could write the literal `pending:<id>` marker into Frappe as a field value.

Full design, case matrix and internals: [`doc/release-2.0/attachments.md`](https://github.com/dhwani-ris/frappe-mobile-sdk/blob/2c4734ec3039c0c1cd195794a3e72867ee2465fc/doc/release-2.0/attachments.md) — new, 803 lines.

**Size:** 71 files, +8,744 / −447 — 30 in `lib/`, 38 tests (28 new), 3 docs. Base branch `develop`.

## What changed

**Pipeline and push gate**
- A resolved `server_file_url` **committed to SQLite** is now the correctness boundary: the url is written back into `docs__<doctype>` in the same transaction as the state transition, so payload assembly reads an ordinary `file_url`. Upload and document push are independent commitments — an uploaded file is never re-uploaded, whether or not the parent ever syncs.
- A parent may push only when **every** attachment is `done`; otherwise it throws `BlockedByUpstream` naming the file and field. `inlinePayload` now throws on an unresolved marker instead of shipping it.
- Terminal/transient split: `AttachmentState.rejected` (oversized, wrong type, not permitted) is never auto-retried; `failed` stays transient and is re-armed next dispatch. Unrecognised errors default to transient.
- Fixed three reproduced paths where a `pending:<id>` marker reached Frappe verbatim (upload OK → parent POST failed → retry; terminal refusal → `blocked` → retry, which *reported success*; any `TimestampMismatchError` via auto-merge recursion, with no user action).
- Offline mode now defers uploads regardless of connectivity — `AttachField` / `ImageField` / `FrappeFormBuilder` / `FieldFactory` gain an `isOfflineMode` callback; `resolvePickedAttachment` gains `offlineModeEnabled`. Both default to previous behaviour.

**Media cache and offline previews (schema v6 → v7)**
- New `media_cache` table + `MediaStore`, keyed by server `file_url` (Frappe dedupes by content hash, so per-document caching would double-store and turn deletion into reference counting). Push-uploaded bytes *move* into the cache; pulled media is cached on first online view. Attach / Attach Image / Image render from disk on a hit; non-image attachments are handed to the external app from the cache. Cache state is non-authoritative — a vanished file is a miss, never an error.
- Storage layout: staged files at `mform_attachments/outbox/<id>/<original filename>`, uploaded bytes at `mform_attachments/cache/<sha256(file_url)>`. Staging is never evicted (it is the only copy); cache always is (re-fetchable).
- **No automatic eviction in this release** — the store grows until logout or an explicit clear.

**Frappe-side correctness**
- `upload_file` was sent `dt` / `dn` / `filename`; `frappe/handler.py` reads `doctype` / `docname` / `file_name` (checked against 16.25.0, 16.26.3, 17.0.0-dev). The old branch had no in-tree caller, so no released behaviour changed.
- The user's filename now survives to the server. Three places discarded it and all three had to change — staging, the enqueue, and the multipart part name in `RestHelper.uploadFile` (load-bearing: Frappe reassigns `filename` from the part whenever a file part is present).

**UI**
- Attachments can be **discarded**, not just replaced (Attach / Attach Image / Image, when editable and non-empty).
- `ImagePickSource` host hook — gallery, camera, or both; read live, global, `null` means both. Attach Image / Image only.
- Pick-time size guard (default 10 MB, matching Frappe's stock `max_file_size`), enforced *before* the durable copy; widgets surface the real limit.
- A value supplied after a field's first build now renders (`hasInteractedByUser` distinguishes "user cleared this" from "never touched").

**Storage APIs and leak fixes**
- `FrappeSDK.mediaStoreUsage()` (staged / cached / reclaimable-now) and `sweepOrphanedMedia()` (exact guards only: unreferenced *and* not staged this session; never throws; deletes nothing if the reference query fails).
- `clearMediaCache()` no longer wipes staged files — it clears `cache/` and the index only. `logout(clearDatabase: true)` is now the only path that clears staged files, where that is the intended security behaviour (previously the store survived logout, leaving one user's photos readable on a shared device).
- Fixed staged-file leaks on every removal path (`hardDeleteLocalMirror`, cancelled-INSERT, tombstone → all route through `deleteForTopParent`), on re-pick, and on a pick that never reached a save.
- Fixed: a completed upload could resurrect a discarded attachment (writeback is now conditional on the column still holding that row's marker); clearing a field left its queued row alive and blocked the document forever; `pending_attachments.size_bytes` / `.mime_type` were never populated; a fresh install wrote `sdk_meta.schema_version` from a drifted literal.

## Also in this PR (not attachments)

- **`mobile_uuid` adoption on pull** — `pull_apply.dart`. A pulled document was stored under a freshly minted `mobile_uuid` even when the server round-tripped this device's own value: the server value was read as a *matching key* only, never stored, and both parent write paths minted unconditionally. Invisible in normal operation (a present local row preserved its uuid), it surfaced whenever the row was gone — `logout(clearDatabase: true)`, reinstall, fresh-device onboarding — giving the same server records brand-new local identities and defeating the `unique` `mobile_uuid` fallback that exists to reconcile an interrupted `server_name` writeback. Precedence is now **local → incoming → mint** through one `resolvePulledMobileUuid` helper used by all four write sites; empty/absent still mints (Desk-origin rows return null/`''`, and it is the local PK). Child behaviour is unchanged, only deduplicated onto the helper.
- **Push-engine ordering tests only** — link-chain order and same-document operation order (no production change).
- **Barrel exports** for `ImagePickSource` / `MediaStoreUsage`, which 2.0.0-beta.2 documented as host-facing but never exported, so the pick-source hook could not be wired at all.
- **Doc corrections** in `attachments.md` and `CHANGELOG.md` (stale "not verified on a device" claims; server-side image shrinkage attributed to the wrong mechanism).

## Breaking / migration

Hosts that only call `FrappeSDK` / `SyncService` are unaffected — `PushEngine` wires the pipeline internally.

- **Schema v6 → v7** adds `media_cache`. Automatic, backfills nothing (an empty cache is just misses); existing `pending_attachments` rows keep working — they store absolute paths.
- `FrappeSDK.mediaStoreSize()` → **`mediaStoreUsage()`** (a single number cannot say how much is reclaimable). Both are unreleased, so nothing shipped against the old name.
- `AttachmentPipeline.uploadPendingForTopParent` → **`resolveForTopParent`**; the constructor now requires `db`; `resolutionMapFor` is new.
- `PendingAttachmentDao.findPendingForTopParent` → **`findUnresolvedForTopParent`**, plus `findAllForTopParent` and `markRejected`.
- `AttachmentPipeline.inlinePayload` throws on an unresolved marker (was a silent passthrough).
- `resolvePickedAttachment` gained `maxBytes` and now rethrows a **terminal** upload failure and discards the staged copy; a transient failure still falls back to staged queueing.
- Staged path layout changed (see above).

## Verification

- `flutter test` on this branch: **2338 tests pass** (exit 0, Flutter 3.44.2). 28 new test files, 10 updated. `test/public_api_surface_test.dart` imports only the public entry point, so it fails to compile if an export is dropped — every other test imports `src/` directly, which is why the missing exports were invisible.
- **Device run** (documented in `attachments.md` §16): Android 8.0.0 / API 26 debug build against a Frappe v16 bench, offline mode on, two documents × two Attach Image fields (camera + gallery). Confirmed: staging keeps the picked filename, **zero** `upload_file` calls before sync, markers never crossed between docs or fields, all four → `done` with real `/private/files/…` urls, a server-side `LIKE 'pending%'` query on both fields returning **empty**, `File` relink correct on all four (reconstructed server-side from `file_url` alone), staged bytes *moved* into `cache/`, and `logout(clearDatabase: true)` removing `mform_attachments/` entirely along with every `docs__*` table and all credentials.
- The `mobile_uuid` fix was reproduced and re-verified on a live bench (bulk pull path): divergence before, byte-for-byte adoption after, a Desk-created record with a null server uuid correctly still minting, children parented to the adopted uuid with zero orphans, 603/603 HTTP 200.
- **Not exercised on hardware** (unit-tested only): offline capture, offline preview from `outbox/`, `BlockedByUpstream` on a gated document, the `failed` and `rejected` states, re-pick/discard reclaim, camera lost-capture recovery, `OpenFilex` handoff.
- Worth knowing (§16.2): with `strip_exif_metadata_from_uploaded_images` enabled, Frappe re-encodes JPEGs through PIL — lossy, GPS EXIF destroyed, dimensions unchanged — so the device cache holds the pre-strip original while the server holds the re-encode. Benign, because the cache is non-authoritative, but a cache hit and a miss can render slightly different bytes for the same url. PNG is untouched (the strip is JPEG-gated). This is a site setting, not SDK behaviour or a Frappe default.

## Known issue (not fixed here)

`ResolveMediaFn` is still not exported, so a host cannot write the signature of a `FieldFactory.createField` override — `FieldFactory` is exported bare and documented as overridable, but the typedef lives in `src/services/media_resolver.dart`, and Dart does not re-export a library's imports. Confirmed against a real consumer app that subclasses `FieldFactory`; it cannot compile without an `implementation_imports` violation. The fix is one export line. **2.0.0 should not ship claiming that type is host-facing until it lands** — `public_api_surface_test.dart` will guard it afterwards.

## Review pointers

- Correctness boundary: `lib/src/sync/attachment_pipeline.dart`, `lib/src/sync/push_engine.dart` (the gate + writeback transaction).
- Identity: `lib/src/sync/pull_apply.dart` (`resolvePulledMobileUuid` precedence).
- Storage lifetime rules: `lib/src/utils/media_store.dart`, `lib/src/utils/attachment_paths.dart`.
- Regression pins: `test/sync/marker_resolution_regression_test.dart`, `test/sync/attachment_multi_field_test.dart`, `test/sync/pull_apply_mobile_uuid_adoption_test.dart`.


deepak-dhwani (changes requested):
# Round 1 — `e665ddf` · Request changes

The attachment pipeline is well-built: the commitment model is right, the crash-point ordering is
explicit, and the storage-lifetime rules are the most careful part of the change. The blocker is not
in the attachments — it is the `mobile_uuid` adoption that rides along with them.

**Scope of this review, stated so it isn't overread.** Static only — no Dart toolchain on my side, so
the reported 2338 passing tests are not reproduced. I read `attachment_pipeline.dart`,
`push_engine.dart` (the gate + writeback), `pull_apply.dart`, `media_store.dart`,
`media_resolver.dart`, `attachment_paths.dart`, `attachment_error_classifier.dart`,
`pending_attachment_dao.dart`, `local_writer.dart`'s producer, the schema/migration, and the barrel.
**I did not review** `image_field.dart` (527 lines, the largest single file), `attach_field.dart`,
`form_screen.dart`, `child_table_field.dart`, the 38 test files, or `attachments.md`. The UI layer is
unreviewed and should not be read as cleared.

---

## 🔴 H1 — `resolvePulledMobileUuid` adopts an unvalidated server string into the PRIMARY KEY

The precedence (local → incoming → mint) is the right fix for the problem described. But `mobile_uuid`
is `TEXT PRIMARY KEY` on both parent and child mirrors (`parent_schema.dart:23`,
`child_schema.dart:19`), and adoption is unconditional on anything except emptiness:

```dart
// pull_apply.dart:64
String resolvePulledMobileUuid({...}) {
  if (localUuid != null && localUuid.isNotEmpty) return localUuid;
  final adopted = incoming?.toString();
  if (adopted != null && adopted.isNotEmpty) return adopted;   // ← no shape, no uniqueness
  return uuidGen.v4();
}
```

So a server-supplied value now becomes the local primary key. Two things follow, and they are
independent.

### (a) Two server documents with the same `mobile_uuid` — different outcomes per path, one of them silent

The four write sites don't agree on conflict policy:

| Site | Path | Insert | Outcome on a duplicate adopted uuid |
|---|---|---|---|
| `:420` parent | sequential | `:462` `batch.insert` (default **abort**) | `batch.commit` (`:559`, no `continueOnError`) throws — page apply fails |
| `:610` parent | **bulk** | `:645` `ConflictAlgorithm.**replace**` | second row **silently overwrites** the first |
| `:530` child | sequential | `:555` default abort | commit throws |
| `:667` child | bulk (`localUuid: null`) | `:693` default abort | commit throws (`:698`) |

The bulk row is the bad one. `mobile_uuid` is the PK and the two rows carry *different* `server_name`s,
so `replace` means **one server document's local mirror is destroyed and replaced by another's, with
no error anywhere**. The document simply isn't on the device. Before this PR that was impossible by
construction — every pulled row minted its own v4.

The sequential and child sites fail loudly instead, but they fail *permanently*: the page throws on
every retry for as long as the server data stays that way.

**The trigger is not exotic.** Frappe's Desk **Duplicate** and **Amend** copy custom field values,
`mobile_uuid` included — so duplicating a mobile-created document in Desk produces a second document
with the same parent *and* child `mobile_uuid`s. A Data Import with a `mobile_uuid` column does the
same. Then any fresh-install / post-`logout(clearDatabase: true)` / new-device pull — precisely the
scenario this fix targets, since that is when adoption fires at all — hits it.

`existingUuids` cannot catch this. It is keyed `server_name IN (this page's names)`
(`:587-592`), so it answers "what uuid does *this* server_name already have", never "is this uuid
already taken by a different one".

### (b) It breaks an invariant the codebase states in prose and relies on for identity

`uuid_rewriter.dart:73-80` says it outright:

> *Frappe server names are never UUID-shaped, and SDK `mobile_uuid`s **always are** — so it catches
> local refs populated by paths that never fire `onIsLocalChanged` (fetch_from, defaults, programmatic
> prefill, back-reference Links). **Relying on the flag alone let those raw UUIDs reach the server,
> which rejects them as unknown names.***

Adopt `HSFM-2026-00042` and `looksLikeMobileUuid` returns false, so the rewrite falls back to
`__is_local` alone — which that same comment documents as insufficient for four specific paths, with
the consequence spelled out. `form_screen.dart` and `link_field.dart` gate on the same proxy.

### Fix

Gate adoption on shape *and* uniqueness — both are cheap, and either alone leaves half the hole:

```dart
String resolvePulledMobileUuid({
  required String? localUuid,
  required Object? incoming,
  required Uuid uuidGen,
  required bool Function(String candidate) isFree,   // new
}) {
  if (localUuid != null && localUuid.isNotEmpty) return localUuid;
  final adopted = incoming?.toString().trim();
  if (adopted != null &&
      adopted.isNotEmpty &&
      looksLikeMobileUuid(adopted) &&   // restores the uuid_rewriter invariant
      isFree(adopted)) {                // no OTHER local row holds it
    return adopted;
  }
  return uuidGen.v4();
}
```

`isFree` wants a real query — `SELECT 1 FROM <table> WHERE mobile_uuid = ?` excluding this
`server_name` — not a reverse scan of `existingUuids`, for the reason above. In the bulk path it can
be batched once per page against the incoming uuid set.

Separately: please make the two parent paths agree on conflict policy. `ConflictAlgorithm.replace` on
a primary key the server can now influence is a silent-overwrite primitive, and having one path abort
and the other replace means the same server data corrupts differently depending on which code path
ran.

Worth adding to `pull_apply_mobile_uuid_adoption_test.dart`: two server rows sharing a non-empty
`mobile_uuid`, asserted through **both** `_applyPageInTxnBulk` and `_applyPageInTxnSequential`, plus a
non-UUID-shaped incoming value asserted to mint rather than adopt.

---

## 🔴 H2 — `ResolveMediaFn` should be exported in this PR, not after it

The description already makes the case, better than I would: `FieldFactory` is exported bare
(`frappe_mobile_sdk.dart:109`) and documented as overridable, `createField` takes
`ResolveMediaFn? mediaResolver` (`field_factory.dart:119`), the typedef lives in
`src/services/media_resolver.dart`, Dart does not re-export a library's imports, and a real consumer
app that subclasses `FieldFactory` cannot compile without an `implementation_imports` violation.
Confirmed all of that.

What doesn't follow is leaving it. This PR adds two barrel exports for exactly this reason, in a
single hunk, and the missing line goes two lines below them:

```dart
export 'src/models/image_pick_source.dart' show ImagePickSource, ImagePickSourceHelpers;
export 'src/models/media_store_usage.dart' show MediaStoreUsage;
+ export 'src/services/media_resolver.dart' show ResolveMediaFn;   // ← this PR
```

The description says *"2.0.0 should not ship claiming that type is host-facing until it lands"* — so
this is a known-broken public API being deferred one line short of fixed, in the file already being
edited. Please add it here and let `public_api_surface_test.dart` start guarding it now.

---

## 🟡 M1 — `cachePathFor` is called with and without `sourcePath`, so the index can name a file that was never written

```dart
// media_store.dart:140 — inside moveToCache, WITH the source
final dest = await cachePathFor(fileUrl, sourcePath: stagedPath);

// attachment_pipeline.dart:205 — recomputed, WITHOUT it
if (moved) cachedPath = await MediaStore.cachePathFor(fileUrl);
```

`cachePathFor` borrows the extension from `sourcePath` when the url has none — its own docstring names
the case (*"a `download_file` proxy url that carries the name in its query"*) and
`media_store_test.dart:79` pins the branch. For such a url the bytes land at `<digest>.jpg` while
`media_cache.local_path` records `<digest>`.

Consequences: `MediaResolver` finds the row, `File(hit.localPath).exists()` is false, treats it as a
miss and re-downloads — every view, forever. And the uploaded original at `<digest>.jpg` becomes
unreachable bytes that **no API reclaims**: `sweepOrphans` walks `outbox/` only (`media_store.dart:296`),
so `clearCache()` is the only thing that will ever remove it. `media_resolver.dart:95` is
self-consistent, so this is the one divergent call site.

**Fix:** pass `sourcePath: p.localPath` at `attachment_pipeline.dart:205` — or better, have
`moveToCache` return the destination it actually used, so the two can't drift again.

---

## 🟡 M2 — every 417 is terminal, which inverts the classifier's own stated policy

`isTerminalAttachmentError` sets out the principle and then contradicts it:

> *Defaults to FALSE for anything unrecognised. A wrongly-transient error costs one retry; a
> wrongly-terminal one strands the user's file with no automatic recovery. **Fail toward retrying.***

```dart
if (error is ValidationException) return true;   // ← every 417
```

417 is not the oversized-file status; it is where **every** `frappe.throw` lands. A site storage-quota
throw, a custom `File` hook, any transient server-side business rule — all become `rejected`, which is
never auto-retried and blocks the parent document until the user re-picks the file. That is the exact
cost the docstring says to avoid, applied to the broadest error class Frappe has.

`_terminalMarkers` already names the two genuinely terminal cases, but it is only consulted on the
flattened-exception path at the bottom — after this branch has returned.

**Fix:** for `ValidationException`, consult `_terminalMarkers` against the message and default to
transient. That reuses the list that already exists and makes the behaviour match the doc.

---

## 🟡 M3 — the marker prefix has four spellings, two of them literals

`kPendingMarkerPrefix` exists (`attachment_paths.dart:30`) and is used at two of the four sites that
must agree byte-for-byte:

| Site | Spelling |
|---|---|
| `local_writer.dart:307` — the **producer** | `'pending:$id'` literal |
| `attachment_pipeline.dart:319-320` — `inlinePayload`, the "loud not silent" backstop | `'pending:'` literal |
| `attachment_paths.dart:19` — `isLocalAttachmentPath`'s `nonLocalPrefixes` | `'pending:'` literal |
| `attachment_pipeline.dart:251` — the writeback `WHERE` | `kPendingMarkerPrefix` ✅ |
| `media_resolver.dart:62` — display resolution | `kPendingMarkerPrefix` ✅ |

Change the constant and the two halves break in opposite directions, both quietly. The writeback
`WHERE` starts matching nothing. And `MediaResolver` stops recognising the marker → falls through to
`isLocalAttachmentPath`, whose literal *still* matches → returns not-local → the marker goes to the
cache/network path, and **a local id is issued as an HTTP request** — the one thing that function's
comment says must never happen (*"Falling through to the network for a malformed marker would turn a
local id into a request path"*).

Nothing is wrong today. But this PR exists because markers reached the server, and the invariant that
prevents it is currently spread across four uncoordinated string literals. Using the constant at all
four sites costs nothing and makes the compiler enforce it.

---

## 🔵 Low

**L1 — `kDefaultSyncBackoff` means a different attempt count in each of its two consumers.**
`attachment_pipeline.dart:154` is `attempt < backoff.length` → 3 attempts, with delays only after
attempts 0 and 1, so the third entry (10s) is never used as a delay. `push_engine.dart:568` is
`attempt <= networkBackoff.length` → 4 attempts. The docstring says *"3 attempts at 2s, 5s, 10s"*,
which is accurate for neither. Trimming the list to `[2s, 5s]` would silently drop the pipeline to two
attempts, because there the list length *is* the attempt count.

**L2 — downloads are unbounded while picks are capped.** A pick is size-guarded at 10 MB before
staging, but `MediaFetchFn` is `Future<List<int>?>` and `MediaResolver.resolve` buffers the whole body
before `writeAsBytes`. Nothing caps a Desk-uploaded attachment, and the verified device is Android
8.0.0 / API 26. Worth a cap on the resolve path, or a streaming fetch.

**L3 — three copies of the fetch-URL logic.** `frappeFileFetchUrl`'s own NOTE records that
`AttachField._fullFileUrl` and `ImageField._fullImageUrl` still hold private copies, deferred to keep
the change small. Reasonable at the time, but `/private/files/` has to route through `download_file`
to carry auth, so a drift between the three is a private-file 404. Now that the shared helper exists,
worth collapsing onto it.

---

## ✅ What holds up

- **The commitment model.** "A `server_file_url` committed to SQLite is the correctness boundary",
  with the four steps numbered by crash point and the explicit note that the pre-commit window
  degrades to a re-upload that Frappe's content-hash dedup absorbs. Upload and document push as
  independent commitments is the right decomposition.
- **The conditional writeback.** `WHERE mobile_uuid = ? AND "<field>" = 'pending:<id>'` with the
  reasoning written down — an unconditional update *is* the same mistake that produced the original
  bug, and matching nothing when the user re-picked or discarded is the correct outcome. This is the
  best single decision in the PR.
- **`isStagedPath` uses canonical containment, not a prefix match.** `p.canonicalize` + `p.isWithin`
  rejects `outbox_old/` and resolves `..`, and the comment says exactly why it matters: callers use it
  to decide whether to **delete**, and a host may point a field at `/sdcard/DCIM/holiday.jpg`. Getting
  this wrong is unrecoverable; it is right, and it is right for a stated reason.
- **Register-before-create in `stageToOutbox`.** Adding to `_stagedThisSession` *before* the file
  exists closes a window where a sweep could see a file with neither a row nor a live-set entry. Also
  the note that the set is empty after restart and those files are then genuinely orphaned — so they
  become sweepable exactly when they should.
- **`deleteOutboxCopy`'s dirname guard.** Only prunes a directory sitting directly under `outbox/`, so
  an arbitrary path can't delete something else.
- **Sweep guards are exact, not heuristic.** No age rule, so a pick sitting in an open form is never at
  risk; and the explicit contract that the caller must not pass an empty `referencedPaths` when the
  query failed, honoured in `sweepOrphanedMedia`.
- **The `clearCache` / `clearAll` boundary**, and the logout fix behind it — the store surviving logout
  and leaving one user's photos readable on a shared device is a real security bug, and the fix puts
  the destructive path where it belongs.
- **Deadlock notes at both pre-resolve sites** (`attachment_pipeline.dart:218`,
  `local_writer.dart:232`): querying the outer `Database` while holding a txn hangs on sqflite's
  non-reentrant lock, and it's called out as a Dart deadlock rather than `SQLITE_BUSY`. That
  distinction saves the next person an afternoon.
- **`public_api_surface_test.dart`** — a test that imports only the public entry point, so a dropped
  export fails to compile. That is the right shape of guard, and the observation that every other test
  imports `src/` directly is exactly why the missing exports were invisible.
- **The migration.** Properly versioned v6→v7, `IF NOT EXISTS` throughout, `ALTER` wrapped for
  re-entry, `schema_version` written in the same transaction, and a re-entrancy test.
- **The verification section names what was *not* exercised on hardware** — offline capture, offline
  preview from `outbox/`, `BlockedByUpstream`, the `failed`/`rejected` states, reclaim, camera
  lost-capture, `OpenFilex` handoff. And §16.2's EXIF-strip note (device cache holds the pre-strip
  original, server holds the re-encode) is the kind of thing most PRs discover in production.

---

## Gate

| Area | Status |
|---|---|
| Attachment commitment model | ✅ |
| Push gate / marker containment | ✅ |
| Writeback atomicity + conditionality | ✅ |
| Staged-file lifetime & reclaim | ✅ |
| Logout / wipe boundary | ✅ |
| Schema migration v6→v7 | ✅ |
| Cache index ↔ filesystem agreement | 🟡 M1 |
| Terminal/transient classification | 🟡 M2 |
| Marker constant discipline | 🟡 M3 |
| **`mobile_uuid` adoption** | 🔴 **H1** |
| Public API surface | 🔴 H2 |
| UI layer (`image_field`, `attach_field`, `form_screen`) | ⚪ not reviewed |

**Requesting changes on H1 and H2.** H2 is one line in a file this PR already edits. H1 is the real
one, and I'd genuinely consider splitting it out: the attachment pipeline is independently verified,
well-tested and ready, while the `mobile_uuid` adoption is a change to primary-key identity that
deserves its own diff, its own collision tests, and its own bench run. It is already flagged as "also
in this PR (not attachments)" — that instinct was right, and the reason it should be separate is that
it is the only part of this change that can silently lose a document.

M1–M3 and L1–L3 are all small and can land in the same round.

Omprakash-48:
Round 1 addressed — 2677711

All nine findings actioned. Suite: 2,364 passing, analyze clean.

H1 — took the shape gate, skipped the uniqueness gate. mobile_uuid is unique: 1 with a real enforced index (a duplicate write fails with MariaDB ERROR 1062, parent and child), so two server docs can't share one and the replace path isn't reachable from adoption. Parent adoption now requires looksLikeMobileUuid — H1(b) was the real defect. Children deliberately excluded: pull_apply_test.dart's orphan-Link regression test uses a non-v4 uuid and asserts adoption, and children have always adopted any non-empty value. Gating them would re-break that.

H2 — done. One line. Test fails to compile without it; the consumer app now builds against the barrel alone.

M1 — done. moveToCache returns the destination it used, so the caller can't recompute it wrong.

M2 — done, structurally rather than by substring. exc_type is available (response.py:52 sets it unconditionally; RestHelper passes the whole body through). Also: oversized never reaches 417 — werkzeug's max_content_length is the same get_max_file_size() (app.py:194), so a 30 MB upload returns 413, already terminal via the 4xx rule. Dropping the blanket rule can't weaken it.

M3 — done. All five sites on the constant.

L1 — docstring only. Both loops are correct for their own purpose; only the shared comment fit neither. Left the loops alone.

L2 — bounded in two places. You're right a cap in resolve() doesn't fix the OOM, so the fetcher now streams and refuses on Content-Length. The resolver cap keeps oversized bodies out of the cache.

L3 — collapsed, after diffing all three and pinning every branch of frappeFileFetchUrl first (it had no coverage).

Also here: child mobile_uuid now ships on the wire, closing the round trip. Verified against a live bench.

Two heads-ups: L2/L3 touch form_screen.dart / attach_field.dart / image_field.dart — the layer you didn't review. And H2 must land with the consumer app change; it currently resolves the SDK from a different ref.

deepak-dhwani (changes requested):
Two new commits (2677711, 982a5f7). All nine round-1 items verified fixed in code.

grep -rn "'pending:" lib/ now returns exactly one hit — the constant's own definition.

Two corrections to my round 1

M2 — you corrected my premise. I accepted the docstring's claim that 417 is how an oversized upload arrives and only argued about breadth. You checked: Frappe sets werkzeug's max_content_length to the same get_max_file_size() (app.py:194), so the HTTP layer returns 413 first — verified with a 30 MB upload. The justification was wrong, not merely over-wide, and classifying on exc_type is better than the narrowing I asked for.

H1 — I was wrong to frame the uniqueness fix as "just add the check." The bulk path has no existing.isEmpty branch: ConflictAlgorithm.replace at :688-691 is its upsert, so it can't become abort, and replace can't distinguish a legitimate same-uuid update from a pathological collision. The only client-side discriminator is the per-row query you declined — so relying on the server-side index is the pragmatic call. My Duplicate/Amend trigger doesn't survive it either: with the index in place, Frappe's own Duplicate fails 1062 first.

The child asymmetry I was going to raise — withdrawn. Both child paths set server_name from cr['name'] (:582, :719), so the reachability mitigation extends to children; and the preservation fixture uses a non-v4 uuid because deployments exist with that shape, so gating would break real installs.

Your find beats mine

Child mobile_uuid never reaching the wire is bigger than anything I raised, and it retro-corrects my H1: with children round-tripping as NULL, the two child write sites in my table were inert in practice, not live. My finding was right about the parent paths and wrong to weight the child ones equally.

One residual — a sentence, not code

The CHANGELOG states the precondition (server-side UNIQUE index) but not the consequence if unmet: on the bulk path replace destroys one document's local mirror silently, while the sequential path aborts loudly. Since this SDK ships to pub.dev standalone and the guarantee is owned by mobile_control, a site provisioning mobile_uuid without unique: 1 gets silent single-document loss. Worth one line beside the existing note.

Why this isn't an approve yet

image_field.dart (527 lines) is still unread, along with child_table_field.dart, the 38 test files, and attachments.md. That's the pick / discard / re-pick / preview lifecycle and the hasInteractedByUser change — the part most likely to hold a widget-lifecycle bug and the part with the least review.

Omprakash-48:
Commit ca69c95 → omprakash/feat/offline-attachment-pipeline (PR 107's head). Two files, comments only, no behaviour change: CHANGELOG.md:30 and resolvePulledMobileUuid's docstring in lib/src/sync/pull_apply.dart. Both now state what happens if the server-side mobile_uuid UNIQUE index is missing.

What the verification changed — I probed the real code instead of reasoning about it, which corrected the reviewer once and me three times:

- Reviewer's "sequential aborts loudly" — true only same-page. Three outcomes, not two: bulk = silent replace; sequential same-page = UNIQUE constraint failed, page rolled back, reported and stalled on both pull paths; sequential across pages = tripwire logs, in-place overwrite.
- My "nothing backfills it" — wrong. The two documents share one row for the install's life and swap on every page carrying either; those later swaps do log.
- My assumption the bulk pre-check would catch it — it doesn't. It counts dirty and local-only rows, so a clean synced row gets replaced silently.
- My "ships to pub.dev standalone" — unverifiable; reworded to what is (no publish_to: none, git-ref dependency, index owned by a separate Frappe app). Also verified mobile_control's "unique": 1 in code rather than inheriting the claim.

deepak-dhwani:
# Round 3 — `ca69c95` · All findings cleared

Every item from rounds 1 and 2 is now closed. Two of your fixes landed better than what I asked for,
one of your own finds is bigger than anything I raised, and you corrected a premise I had wrong.

**Basis:** static only — no Dart toolchain here, so the reported 2,364 passing tests are not
reproduced. `e665ddf..ca69c95` is 27 files, +866 / −119.

---

## Round-1 items — all verified

| ID | Finding | Fix | |
|---|---|---|---|
| H1 | Unvalidated server string into the PK | parent shape gate; uniqueness declined with evidence | ✅ |
| H2 | `ResolveMediaFn` unexported | barrel `:38` + compile-time guard | ✅ |
| M1 | `cachePathFor` called with and without `sourcePath` | `moveToCache` returns `String?` | ✅ |
| M2 | Every 417 terminal | classified on `exc_type`, not status | ✅ better |
| M3 | Four spellings of the marker prefix | constant at all five sites | ✅ |
| L1 | `kDefaultSyncBackoff` attempt count | per-consumer table in the docstring | ✅ |
| L2 | Downloads unbounded | streamed, bounded twice | ✅ |
| L3 | Three copies of the fetch-URL logic | both private copies delegate | ✅ |
| R2 | UNIQUE-index precondition not stated | three outcomes enumerated and verified | ✅ better |

**H2** — the guard test is the right shape: it *annotates* rather than infers, then assigns a
conforming closure, so it breaks both if the export is dropped and if the typedef drifts.

**M1** — returning the destination beats threading `sourcePath` through. The two paths can no longer
disagree, which was the actual defect.

**M3** — `grep -rn "'pending:" lib/` now returns exactly one hit: the constant's own definition.

**L1** — better than making the loops agree, for the reason you give: forcing agreement would change
push retry counts nobody asked for. The list genuinely is two contracts; documenting both is the fix.

**L2** — `Content-Length` refusal before the first chunk, mid-stream re-check for chunked responses
that under-report, `maxFetchBytes` keeping an oversized body out of the cache, and
`client?.close()` in a `finally`. The docstring is honest that the second stage bounds the *cache*,
not the allocation. The 25 MB download ceiling being deliberately above the 10 MB pick ceiling is
right and justified where a reader will find it.

---

## The round-2 residual — closed, and more thoroughly than asked

I asked for one sentence on the consequence of the UNIQUE precondition being unmet. What landed
enumerates three outcomes and marks which is silent, verified rather than reasoned. I checked the two
load-bearing claims:

- **"the pre-check counts locally-dirty and local-only rows, so a clean `synced` row is replaced
  without objection"** — confirmed. The bulk pre-check is a single `COUNT` over
  `sync_status IN (_locallyDirtyStatuses + 'deleted')`, so a `synced` row is not counted.
- **"both pull paths stop the doctype without advancing its cursor … reported, not lost"** —
  confirmed at `pull_engine.dart:423`: *"Mid-pull failure: do NOT persist cursor"*, and the
  `setLastOkCursor` write at `:406` sits before it on the success path only.

Narrowing the silent window to *the bulk page itself* — because every later flip does log through the
clash tripwire once the fallback query starts matching — is a sharper statement than my finding was,
and it is the difference between "silently wrong forever" and "one unannounced event". Noting that
children always take the loud path, and closing with *"do not relax the server-side index expecting a
client-side net"*, puts the constraint where the next person will hit it.

I also accept that I framed H1's fix wrongly. `ConflictAlgorithm.replace` at the bulk parent insert
**is** that path's upsert — it has no `existing.isEmpty` branch — so it cannot simply become `abort`,
and `replace` cannot distinguish a legitimate same-uuid update from a pathological
same-uuid/different-`server_name` collision. The only client-side discriminator is the per-row query
you declined, so relying on the server-side index is the pragmatic call rather than a shortcut. My
Duplicate/Amend trigger doesn't survive it either: with the index in place, Frappe's own Duplicate
fails 1062 before a second such document can exist.

**The child asymmetry — objection withdrawn.** I was going to raise that this PR makes child adoption
newly live (it does, since children previously round-tripped as NULL) and that leaving it ungated keeps
the blind spot open for the case you cite as load-bearing. Two things you documented close it: both
child write paths set `server_name` from `cr['name']` (`:582`, `:719`), so the reachability mitigation
extends to children; and the preservation fixture uses a deliberately non-v4 uuid *because deployments
exist with that shape*, so gating would break real installs. History over taste, as you put it.

---

## Two corrections to my round 1

**M2 — you corrected my premise, and the fix is better for it.** I accepted the docstring's claim that
417 is how an oversized upload arrives and only argued about breadth. You checked: Frappe sets
werkzeug's `max_content_length` to the same `get_max_file_size()` (`app.py:194`), so the HTTP layer
returns **413** before `MaxFileSizeReachedError` can be raised — confirmed with a 30 MB upload on a
live bench. The branch's stated justification was wrong, not merely over-wide. Classifying on
`exc_type` from the decoded body (v2's `errors[].type` read too, substring check demoted to a fallback
for a body carrying no type) is structured rather than a guess, and 413 lands on the 4xx rule where it
belongs.

**Child `mobile_uuid` never reaching the wire is a better catch than anything in my round 1** — and it
retro-corrects my H1. `PayloadAssembler` re-added the parent's uuid after the system-column strip and
the child loop never put it back, so, as your CHANGELOG puts it, *"the pull path's child adopt branch
could only ever fire for a value someone had set in Desk."* My "four write sites" table therefore
overstated the child half: those two sites were inert in practice, not live. The finding was right
about the parent paths and wrong to weight the child ones equally.

The three deliberate details all check out: the child's own uuid rather than `row.mobileUuid` (which
would collide on the second child), omitted-when-blank rather than `''` (MariaDB permits many NULLs
but not many empty strings in a unique index), and the `set_only_once` check confirming re-sends are
safe. No backfill, self-healing on next push, is the right migration posture. Retiring the stale
`ThreeWayMerge` note rather than leaving a comment claiming work already done is the kind of cleanup
that usually gets skipped.

---

## Why this stays `CHANGES_REQUESTED` — and it is not a finding

**Nothing I have reviewed is unresolved.** The gate is held for one reason only: a section of this PR
that I have not read.

Unchanged from round 1:

- **`image_field.dart`** — 527 changed lines, the largest single file in the diff, still unread.
- **`attach_field.dart`**, **`child_table_field.dart`** — read only where later rounds touched them
  (the `frappeFileFetchUrl` delegation).
- **`form_screen.dart`** — read only `_fetchMediaBytes`.
- **The 38 test files** and **`attachments.md`** (803 lines).

That covers the pick / discard / re-pick / preview lifecycle and the `hasInteractedByUser` change —
the part of this PR most likely to hold a widget-lifecycle or state bug, and the part with the least
review. It is also the part your own verification section lists as unit-tested only: offline capture,
offline preview from `outbox/`, re-pick/discard reclaim, camera lost-capture recovery, `OpenFilex`
handoff.

So: no changes are being requested of you. Either I do that pass next, or you tell me to approve on
the strength of the device run and the 2,364-test suite and I'll do that with the gap stated in the
approval. Your call — please don't sit blocked waiting on me if you'd rather ship it.

---

## Gate

| Area | R1 | R2 | R3 |
|---|---|---|---|
| Attachment commitment model | ✅ | ✅ | ✅ |
| Push gate / marker containment | ✅ | ✅ | ✅ |
| Cache index ↔ filesystem | 🟡 M1 | ✅ | ✅ |
| Terminal / transient classification | 🟡 M2 | ✅ | ✅ |
| Download bounds | 🟡 L2 | ✅ | ✅ |
| Fetch-URL duplication | 🟡 L3 | ✅ | ✅ |
| `mobile_uuid` shape invariant | 🔴 H1 | ✅ | ✅ |
| `mobile_uuid` uniqueness | 🔴 H1 | 🟡 undocumented | ✅ three outcomes stated |
| Child `mobile_uuid` round trip | ⚪ missed | ✅ your find | ✅ |
| Public API surface | 🔴 H2 | ✅ | ✅ |
| UI layer | ⚪ not reviewed | ⚪ | ⚪ **still not reviewed** |

Worth keeping from these three rounds: every round-1 finding was two call sites that had to agree and
didn't — a path computed with and without an argument, a prefix spelled two ways, a backoff list read
as two contracts, three copies of a URL builder, two conflict policies. Four of the fixes deleted the
second copy rather than syncing it (`moveToCache` returning its own destination, the constant, the two
delegations), and the one that couldn't be deduplicated got its asymmetry written down with the
failure mode enumerated. That is the right disposition of the class, and it is why I would not expect
it to recur here.

Omprakash-48:
Did the UI pass myself rather than have you approve blind — it turned up one real
defect, so don't approve with the gap stated. Fixed in `93cfb7a`.

**Discard/re-pick after a save deleted the bytes behind a committed queue row.**
`discardReplacedValue`'s "a raw staged path was never saved" reasoning is true of the
column, not of the value a form hands over: `LocalWriter` rewrites the column to
`pending:<id>`, nothing patches that back into an open form, and `hasInteractedByUser`
pins the raw path. Confirmed empirically — after `didChange('/…/outbox/u1/photo.jpg')`,
a rebuild supplying `pending:42` leaves `fieldState.value` at the raw path. Deleting it
strands the queued upload: non-terminal error, backoff exhausted, document blocked, no
auto-retry.

Fixed with a `ReclaimAttachmentFn` seam wired to
`OfflineRepository.reclaimDiscardedAttachment`, which refuses a file
`referencedLocalPaths()` still lists. Default stays `MediaStore.discardValue`. Wired at
every hop including both child-table paths, and each hop mutation-tested.

Also reproduced what you couldn't: `flutter test` +2376 (2364 + 12 new), analyze clean,
format 0 changed.

**New finding, raising not fixing:** this branch added five named params to
`createField` (`isOnline`, `pendingAttachmentPaths`, `mediaResolver`, `isOfflineMode`,
`imagePickSource`) in `df19398`/`28b0f59` — none exist at `v2.0.0-beta.2`. Each breaks
every subclass override at compile time. CHANGELOG:240 says the signature is frozen and
CHANGELOG:38 says a subclassing consumer analyses clean; both can't hold. Mine went in
as factory instance state instead. Move those five the same way, or correct the two
claims — your call.

deepak-dhwani (changes requested):
First round with a toolchain. I installed Flutter 3.35.7 (the pubspec floor) and actually ran the suite — rounds 1–3 were all static:

flutter test    → 2376 passed, exit 0     (reproduces the +2376 claim exactly)
flutter analyze → clean for lib/          (2 pre-existing, neither from this PR)

Every finding below was then confirmed by running probe code, not by reading.

What I found

🔴 H1 — Attach fields download the whole file on render, not on view. MediaResolveBuilder._start() runs from initState, and MediaResolver.resolve fetches on a cache miss. Probe: an AttachField holding /private/files/big-survey-report.pdf issued the fetch with zero taps. For ImageField that's fine (Image.network fetched anyway); for a PDF there's no render-time need at all. Opening a form now pulls every attachment on it, up to 25 MB each — on Android 8 / metered rural connections. And attachments.md:28 bills the read path as "lazy download" while :657 says "No background prefetch — cached on first view." Rendering a form isn't viewing a PDF.

🔴 H2 — Omprakash's own find, confirmed. createField gained five named params on this branch, so every FieldFactory subclass breaks at compile time. CHANGELOG:240 says the signature "should be treated as frozen"; form_builder.dart:431-442 repeats it verbatim in the same file whose :1409 and :2267 call sites pass the new arguments. One thing that strengthens his case: field_factory_override_compat_test cannot catch this — its "prior signature" baseline is ca69c95, four commits after the five params landed.

🟡 M1 — cachePathFor builds the cache filename from an unsanitised URL. p.extension takes the query string too. Probe, with the multi_cloud_storage proxy URL shape the code names by vendor:
basename = <sha256>.abcdefghijklmnop
with a 300-char signed key → errno 63, File name too long
The catch-all swallows it → no media_cache row → silent permanent cache miss, re-download forever. attach_field.dart:82 already has the guard (_safeExtension) 400 lines away.

🟡 M2 — the pick path reclaims fieldState.value, the discard path reclaims the rendered value; probe shows them diverging (null vs the staged path) on an async document load. 🟡 M3 — form_builder.dart:451 clobbers a host factory's reclaimAttachment with the unsafe default on every build, while :560 uses ??= for the sibling hooks. 🟡 M4 — frappe_attachments/ temp cache is outside clearAll, sweepOrphanedMedia, mediaStoreUsage and logout(clearDatabase: true), which matters because this PR sells the logout wipe as a security fix. 🟡 M5 — _fetchMediaBytes has a connect timeout but no stall timeout; its sibling in attach_field.dart:578 does. Plus three lows.

93cfb7a itself is correct — I have nothing to add to it. Nothing here touches the pipeline; rounds 1–3 held up under an actual suite run.

Omprakash-48:
Fixed in 1ee7dfa. All findings confirmed by running probe code first.

**H2 — createField break.** Confirmed. The five params are now instance state.
Verified against a real app with a `FieldFactory` subclass: `invalid_override`
before, 0 errors after. Your point about the compat test was the key one — its
baseline was ca69c95, *after* the params landed, so it could never have caught
this. Now pins the published v2.0.0-beta.2 signature.

**H1 — eager download.** Confirmed, and you were more precise than me. I assumed
"defer for non-image"; wrong — `AttachField` renders only a *button*, so the
eager resolve was unnecessary for all of it, not just PDFs. Now resolves on tap.
`ImageField` stays eager (its preview paints from the local file). `:28` corrected.

**M1 — cache filename.** Confirmed: 309-char extension → 373-char filename vs
`NAME_MAX` 255. Extensions now rejected, not truncated.

**M2 — reclaim mismatch.** Confirmed. Both paths now share one
`liveAttachmentValue` helper — the bug was two copies drifting, so one copy is
the fix.

**M3 — clobbered host factory.** Confirmed, and worse than reported: my own H2
fix had widened it from 1 field to 6 (the five went to `null`). `??=` isn't the
fix — it also blocks the update case. Now assigned only when supplied.

**M4 — temp cache.** Confirmed, and not a subdir issue: it's under
`getTemporaryDirectory()` while the store is under
`getApplicationDocumentsDirectory()`. Different roots. Kept out of
`sweepOrphans` deliberately; also reclaimed by `clearMediaCache()`.

**M5 — no stall timeout.** Confirmed. Per-chunk bound, so slow large downloads
still finish.

**Found while auditing:** three cases of two independent constructions of the
same path — the same class as the earlier `moveToCache`/`resolve` bug.
`getTemporaryDirectory()` is now resolved in one place. The capture marker also
survived logout; fixed, path unchanged.

**Needs sign-off (not just review):**
- H1 changes fetch timing — a form no longer warms the cache, so a first tap
  offline misses. This is the one item with schedule risk.
- `MediaStoreUsage.totalBytes` now includes `viewerTempBytes`, so hosts showing
  it will see a bigger number.

**Not covered by tests:** H1's miss fallback, M2's pick call site
(`FilePicker.pickFiles()` is static, no seam), M5's `send` timeout.

**Upgrade note:** run `flutter pub get`, don't just bump the ref. `open_filex`
has been a dep since v2.0.0-beta.2; two apps on older refs fail to compile on
their existing lockfiles. Breaks `--offline` and vendored pub caches.

analyze clean · 2413 tests pass · format clean · example/ clean. Both consumer
apps: 0 errors, results identical to their own pinned SDK.

deepak-dhwani (approved):
Done, and it's ready to merge

PR #110, opened 2026-08-26
fix(ui): keep single-option preselect off Frappe reserved fields

Omprakash-48:
The Select/Link single-option preselect fired for every field, including the ones Frappe owns. Verified against frappe/develop rather than assumed:

  - amended_from  Link -> self,         read_only:1, NOT hidden  (renders)
  - auto_repeat   Link -> Auto Repeat,  read_only:1, NOT hidden  (renders)
  - parent_<x>    Link -> self,         neither hidden nor read-only
  - old_parent    Link -> self,         hidden:1   (never reaches a widget)
  - naming_series Select, assigned server-side at insert
  - name/owner/modified_by are std_fields pseudo-docfields; Frappe's
    `restricted` tuple refuses them as real DocFields

parent_<x> is the dangerous one: editable and visible on every is_tree DocType, so a tree with one existing node silently acquired a parent the user never picked. naming_series matters too -- get_default_naming_series returns the first TRUTHY option and notes "Empty strings are used to avoid populating forms by default", so the server, not the client, picks it.

The nestedset parent fieldname is frappe.scrub(f"Parent {self.name}"). frappe.scrub is only .replace(" ","_").replace("-","_").lower(), which is WEAKER than normalizeDoctypeTableName -- that one also collapses apostrophes, so reusing it would mispredict parent_item's_group as parent_item_s_group and miss the field. Ported separately as frappeScrub.

Also fixes three preselect defects found alongside:

  - A single-option multi-select could not be unchecked. The preselect lived in build(), so the clear (which emits '') re-fired it on the next frame. Now gated on value == null, distinguishing "never set" from "explicitly cleared". With reqd:1 this had produced a required error while _formData still held the value, and formValues.addAll(_formData) let that win on save.
  - Preselect ignored readOnly/enabled, which is what actually reaches amended_from and auto_repeat.
  - Duplicate options crashed FormBuilderDropdown ("There should be exactly one item with [DropdownButton]'s value") and defeated preselect by making the count 2. Both Select options and LinkField's host-supplied static list are now deduped, order-preserving.

allowPreselect defaults to true so a host constructing these widgets directly is unaffected. FieldFactory carries `doctype` as a mutable instance field rather than a createField parameter, per the extension contract documented on the class -- the override compat test still passes.

Not addressed here: the dirty-on-open baseline is only partly improved. That is a dirty-baseline issue, not a preselect one (a Table MultiSelect field or any defaultValue dirties an untouched doc the same way), so it needs its own design call.

Tests: 40 new across 6 files; 2453/2453 pass, analyze and format clean.


deepak-dhwani (changes requested):
# PR 110 — `b617791` · The code is right. Two documentation items before merge.

Thanks for splitting it — and it split perfectly: same SHA, no cherry-pick, no conflict, because #107
merged first and this sits directly on `1ee7dfa`.

Because the SHA is unchanged, everything I verified in the #107 thread applies here without re-checking.
Recording it in this PR so it lives where the change does.

---

## Verified

**The Frappe research holds — all eight claims, checked against `frappe/develop` independently.**

| Claim | Source | |
|---|---|---|
| `scrub` = `.replace(" ","_").replace("-","_").lower()` | `utils/data.py:3126` | ✅ |
| `restricted` contains `name`, `owner`, `modified_by` | `doctype.py:480-492` — `owner` at `:484` | ✅ |
| `name`/`owner`/`modified_by` are `std_fields` Links | `model/__init__.py:142-158` | ✅ |
| `amended_from` — Link → self, `read_only:1`, **not** hidden | `doctype.py:963-975` | ✅ |
| `auto_repeat` — Link → `Auto Repeat`, `read_only:1`, not hidden | `doctype.py:984-993` | ✅ |
| `old_parent` — Link → self, `hidden:1` | `add_nestedset_fields` | ✅ |
| parent field = `frappe.scrub(f"Parent {self.name}")`, neither flag | `add_nestedset_fields` — literally `parent_field_label = f"Parent {self.name}"` | ✅ |
| `get_default_naming_series` returns the first **truthy** option | `naming.py:501-509`, with the comment you quote verbatim | ✅ |

I went looking for the one I expected to be wrong — `owner` in the `restricted` tuple — and it is there.

**The guard set has no holes.** Frappe's other reserved names (`parent`, `parentfield`, `parenttype`,
`creation`, `modified`, `docstatus`, `idx`) are none of them Link or Select in `std_fields`, so none can
reach a preselect. Complete for the fieldtypes the guard covers.

**`frappeScrub` is justified.** `table_name.dart:24-27` collapses `[^a-z0-9]+`, so reusing it really
would mispredict. Probed: `isFrappeReservedField("parent_item's_group", doctype: "Item's Group")` → true;
`isFrappeReservedField("parent_company")` → false, so an ordinary user Link on a non-tree DocType is
untouched.

`parent_<x>` being the dangerous one is the right call — the only one of the seven that is both editable
and visible, so the only one where a user could ship a wrong value without ever seeing a field.

**`doctype` as instance state** is the `createField` contract applied correctly, and assigning it
*unconditionally* in `_configureFieldFactoryForMeta` is right for the reason established in #107:
meta-derived values are unconditional, host choices are guarded. You drew that line correctly without
being asked.

**I tested the merge result, not just the branch.** Worth flagging: this PR's parent is `1ee7dfa`, not
develop's tip, so `develop..pr110` is one commit but `pr110..develop` is seven — the CI run on this branch
tested a tree that is not what lands. I merged locally and ran it:

```
flutter analyze  → clean for lib/  (2 pre-existing, neither yours)
flutter test     → 2467 passed, 3 skipped, exit 0
```

Clean. The 3 skips are pre-existing from develop (`3905592`, in
`form_builder_reactive_fetchfrom_link_test.dart`) and this PR does not touch that file. Still worth
rebasing onto develop so CI exercises the tree that actually merges.

---

## 🟡 M1 — the `value == null` gate does more than the description says

Carried unchanged from #107, since the commit is unchanged. **This is a documentation fix, not a code
one — I am not asking you to change the behaviour.**

The description presents `value == null` as the multi-select uncheck fix. It gates *both* branches, and
what it actually distinguishes is *absent key* from *empty string* — not "cleared" from "never touched".
Those coincide only if `''` can only arrive from a clear, and it cannot: Frappe stores an unset `Select`
as `varchar NOT NULL DEFAULT ''` and returns `""`, the pull writes it verbatim, and
`form_builder.dart:560` does `_formData.addAll(widget.initialData ?? {})` with no normalisation between.
Probed against the widget:

```
value = null → emitted [Only]    ← new document, key absent
value = ''   → emitted []        ← pulled document, single-select
value = ''   → emitted []        ← pulled document, multi-select
```

So the effective rule is now **preselect fires only for a key absent from form data** — in practice, only
on documents this device created. A pulled record with a one-option Select stays empty, and if it is
`reqd` the user must pick the only choice by hand before the form will save.

**The behaviour is right.** Auto-filling a *synced* record writes a value the user never chose and dirties
a document on open — the same category as the bug this PR fixes, and your own note about the dirty
baseline says you see it.

**And I withdraw the alternative I offered in the #107 thread.** I suggested gating on
`!formData.containsKey(fieldname)` if you wanted the narrow version. Develop now contains `843b86b`
*"seed std eval fields on a non-null value, not key presence"*, whose whole point is that keying on
`containsKey` is the trap — an explicit `{'docstatus': null}` is *"how an unsaved doc is routinely
assembled"*. `value == null` is not just acceptable, it matches the convention this repo just fixed a
data-loss bug to establish. Keep it.

What is missing is only the naming. The single thing pinning this population is a test titled
**"an explicitly cleared value is not re-preselected on mount"**, which describes one of its two callers —
and the more common one, a pulled document, appears nowhere in the commit message, the test suite or the
PR description. A field team notices within a day of rollout that a form stopped filling something in.
Rename the test to cover the pulled case, and say it in the description.

---

## 🟡 M2 — no CHANGELOG entry at all

`b617791` does not touch `CHANGELOG.md`, and `grep -in "preselect\|naming_series\|amended_from\|nestedset"`
over the file returns nothing. That is conspicuous next to #107, where every fix got a paragraph — and
this is the change with the **widest blast radius of anything merged this week**: single-option preselect
for `Select` and `Link`, on every DocType, in every deployment.

It needs an entry covering all four parts, with the fourth marked as a behaviour change rather than a fix:

1. the reserved-field guard (fix),
2. the `readOnly` / `enabled` gate (fix),
3. duplicate-option dedup (fix — it was a crash),
4. the population change (**behaviour**, per M1).

---

## Verdict

The code needs no changes and I would merge it as it stands. I am holding only on the two documentation
items, because both are the kind that never get written after a merge, and because #4 above changes what
users see in every deployment while currently appearing in no changelog and no test title.

Both are small. Rename one test, add one CHANGELOG entry, add a paragraph to the description — and rebase
onto develop while you are there so CI tests the tree that lands. Re-approve on sight.

Omprakash-48:
## Preselect population: what the `value == null` gate actually changes

Adding this because the PR description previously framed `value == null` as the
multi-select uncheck fix alone. It gates more than that, and the second caller is
the common one.

The gate fires the preselect only when the form holds **no entry** for the field —
not merely when the entry is unusable. That has two callers:

1. **An explicit clear** stores `''` (multi-select emits `_listToValue([])`). This
   is the bug the PR was written for: the preselect lived in `build()`, so the
   clear re-fired it on the next frame and pushed the value back, while
   `FormBuilderCheckboxGroup` — whose `ValueKey` had not changed — stayed visibly
   unchecked.

2. **A pulled document** also arrives holding `''`, and this is the far more common
   caller. Frappe stores an unset Select as `varchar NOT NULL DEFAULT ''` and
   returns `""`; the pull writes it verbatim and `form_builder.dart`'s
   `_formData.addAll(widget.initialData ?? {})` normalises nothing in between.

So the effective rule is now: **a single-option `Select` is auto-filled only on a
document this device created.** A synced record with a one-option Select stays
empty, and where that field is `reqd: 1` the user must pick the sole choice by hand
before the form will save.

That is deliberate — auto-filling a synced record writes a value the user never
chose and dirties an untouched document on open, which is the same category as the
bug this PR fixes. It is called out explicitly because it is the widest-reaching
part of the change (every DocType, every deployment) and because the symptom a
field team meets is "a form stopped filling something in", which does not look like
a fix.

Keying on `value == null` rather than `!formData.containsKey(fieldname)` also
matches the convention `843b86b` established: an unsaved document is routinely
assembled with explicit `null` entries, so key presence is the trap.

### Correction to the review's scope for this item

The review's probe table generalised from `SelectField` to both widgets. It does not
extend to `LinkField`, and this PR left `LinkField`'s gate untouched:

| widget | branch | gate | `value == ''` |
|---|---|---|---|
| `SelectField` | single + multi | `value == null` | **suppressed** |
| `LinkField` | static options | `validInitialValue == null \|\| validInitialValue.isEmpty` | **still preselects** |
| `LinkField` | service-loaded | `!hasValidSelection` | **still preselects** |

Verified by test, not by reading — see `link_field_preselect_test.dart`, group
"an empty stored value". So a single-option **Link** *does* still auto-fill over a
pulled document's `''`, where the equivalent Select no longer does.

Not aligned in this PR, on purpose. The clear/re-fire loop that forced Select's
stricter gate is specific to the multi-select checkbox path, which `LinkField` has
no equivalent of; matching the two is a behaviour change that owes its own evidence
and migration note rather than a ride-along here. The asymmetry is now pinned by
test and cross-referenced from the doc comment on `SelectField._canPreselect`, so
the next person to touch either widget meets it deliberately. Happy to open a
follow-up if you want them aligned.

The reserved-field guard still wins over the Link behaviour — `parent_<doctype>` on
a one-node tree stays untouched even though its stored value is the same `''`. That
is covered by a third test in the same group.

ankitDhwani:
Both commits from this branch have been carried into #116 with `git cherry-pick`, so authorship is preserved — `17f631c` and `e0239c7` there are byte-identical to `52c0a98` and `7a13137` here, still authored by @Omprakash-48.

Reason for combining rather than merging both: #116 widens the `file_picker` range, and that turned out to have introduced a live regression in `attach_field.dart` (12.0.0 removed `allowMultiple`, so `pickFiles()` is always multi-select and the call site's `files.single` threw on a two-file selection). Both PRs touch the form-field layer and the same `## [Unreleased]` block, so landing them separately meant two CI runs and a conflict for whichever merged second. One review, one run.

Nothing was changed in the preselect work itself. One thing **was** corrected around it: the CHANGELOG entries here sit under `## [2.0.0-beta.3] - 2026-09-17`, which now reads as a claim that they shipped in that release. They did not — `v2.0.0-beta.3` does not contain `lib/src/utils/frappe_reserved_fields.dart`. That was correct when written on 2026-09-01, when beta.3 was still the unreleased section at the top of the file; the release simply moved underneath it. The four bullets were moved verbatim into `[Unreleased]`, with the deliberate behaviour change (single-option `Select` no longer auto-filling on a pulled document) under `### Changed` rather than `### Fixed`.

Combined suite on #116: 2659 pass, 3 skipped, 0 fail; analyze and format clean.

This PR can be closed in favour of #116 whenever you're happy with that — leaving it to you rather than closing someone else's PR. If you'd rather it land on its own instead, say so and I'll drop the two commits from #116 and rebase onto it once it merges.

One question on the work itself, since I'm now carrying it: `frappeScrub` handles the `is_tree` parent for a DocType whose name contains an apostrophe, which `normalizeDoctypeTableName` would have mispredicted. Was that from a real DocType you hit, or defensive? Asking because if it's real it's worth a test fixture named after it.

PR #111, opened 2026-09-12
feat(ui): child table grid columns, row guidance hook, and a rebuild-loop fix

ankitDhwani:
Brings the child table field up to grid parity so a host app no longer has to
reimplement it, and fixes the defect that made hosts fork it in the first place.

## The blocker this removes

`_ChildTableSheet` rebuilt itself on **every** submit-callback registration.
Hosts register that callback from inside their own `build` — it closes over form
state, so it is recreated each build — which meant `setState` re-entered
`formBuilder`, which registered again, which scheduled another `setState`. A
frame was always pending: the sheet never idled on device, and `pumpAndSettle`
never returned in tests.

Now the sheet rebuilds only on the unregistered -> registered transition, which
is the one moment the action button must flip from disabled to enabled.
Regression test verified to fail with the fix reverted.

## Child table changes

- Render each `in_list_view` column the child doctype declares as a labelled
  cell, with Link cells resolved to document titles through an injected
  resolver. Falls back to the previous single-title tile when none are declared.
- Render `Check` as Yes/No, `Date`/`Datetime` as `dd/mm/yyyy`, and `Attach` as a
  filename alone — stripping a query string, never printing an absolute
  on-device path, and reading a `pending:<id>` marker as human text.
- Bound both halves of a cell line; an unbounded label wrapped to three lines
  and an unbounded Small Text value let a ten-column child eat a third of the
  screen per row.
- Add a `rowNoticeBuilder` hook so a host can render per-doctype guidance above
  a row form — typically for a row not yet on the server.
- Stamp the child doctype onto every emitted row, on both the add and edit paths.
- Title a row from the doctype's declared `title_field` before falling back to
  the conventional name columns and then field order.
- Export `ChildTableField`, `ChildRowNoticeBuilder`, `preserveChildIdentity` and
  the cell helpers on the public barrel, so a host with its own field factory can
  construct and reuse them.

New pure resolver in `child_table_cells.dart` with unit tests; widget tests
extended for columns, the notice hook, sheet insets, doctype stamping and the
rebuild loop.

## Scope note for reviewers

This branch also carries previously unmerged work that the child-table commits
sit on top of — the offline attachment pipeline, mobile creation capture, and
assorted meta/query fixes. They are included because the grid work builds
directly on them, not because they are part of this change. Review the
child-table commits (`df0c0cf` onward) as the substance of this PR.

## Verification

`flutter test` — **2545 passed / 3 skipped**. `flutter analyze lib test` — clean.
Not exercised on a device in this branch.


deepak-dhwani (changes requested):
## PR #111 — child table grid columns, row guidance hook, rebuild-loop fix — Round 1 Review

Reviewed the full branch diff against `develop` (+6,777 / −503, 60 files), not just
`df0c0cf` onward — all of it merges. Findings below are ordered by severity, and I've
marked which stream each belongs to.

---

### 🔴 HIGH H1 — A `mobile_uuid` AND-clause is dropped, and the query goes out unfiltered

`lib/src/query/unified_resolver.dart` — `_toServerFilters` (carried work)

`mobile_uuid` is `continue`d before the `systemParentColumnNames` → `return null`
branch, so it is dropped from AND filters as well as OR filters. For
`filters: [['mobile_uuid','=','u-1']]` the method returns `[]`, not null — the
`serverFilters == null` guard doesn't fire, and the `orFilters.isNotEmpty &&
serverOrFilters.isEmpty` guard doesn't apply because `orFilters` is empty. So
`_onlinePassthrough` runs with **no filters at all** and answers with page 1 of the
entire doctype. The caller asked for one specific unpushed local row.

This is the exact outcome the method's own doc comment says it refuses:
*"silently dropping an AND clause would send a BROADER query than the caller asked
for."* The OR side is guarded **and tested** (`sends NOTHING when an OR list
sanitizes away entirely`); the AND side has neither.

No current in-SDK caller hits it — `LinkOptionService._resolveLinkTitle` uses
`orFilters` and its `server_name` clause survives sanitisation. But `UnifiedResolver`
is exported from the public barrel, so a host composing its own local filters lands
straight in it.

**Fix**: mirror the OR guard — `if (filters.isNotEmpty && serverFilters.isEmpty) return _emptyResult;`
(and the equivalent early `return` in `_scheduleBackgroundRefresh`), plus a test in
the same file as the OR one.

---

### 🔴 HIGH H2 — `ChildTableFormBuilder` is a breaking API change with no BREAKING note

`lib/src/ui/widgets/fields/child_table_field.dart:33-38`

The typedef gains `bool readOnly`. In Dart a function type carrying a named parameter
is only satisfied by a function that declares it, so every host-supplied
`(childMeta, initialData, onSubmit, {registerSubmit})` stops compiling. The proof is
in this diff: both of the SDK's own call sites in `form_builder.dart` had to be
rewritten to add `bool readOnly = false`.

`ChildTableFormBuilder` is exported from the public barrel and is *the* documented
host extension point for child rows. This repo already has the convention for this —
the `DependsOnEvaluator` entry in the same CHANGELOG is a model of it — but there is
no `### Changed — BREAKING` entry here, and the PR description doesn't mention it.

**Fix**: add the BREAKING CHANGELOG entry with the one-line migration
(`{registerSubmit}` → `{registerSubmit, bool readOnly = false}`) and call it out in
the PR body.

---

### 🔴 HIGH H3 — `resolveLinkTitle` and `rowNoticeBuilder` are unreachable through the SDK's own render path

`lib/src/ui/widgets/fields/field_factory.dart:502`

`_TableFieldBase.build()` constructs `ChildTableField` and threads `creationCapture`
through — but not `resolveLinkTitle`, and not `rowNoticeBuilder`. Grepping the branch,
the only occurrences of either outside `child_table_field.dart` are in tests.

So on the path every host actually uses — `FrappeFormBuilder` → `FieldFactory` →
`ChildTableField` — both are permanently null. Link cells render raw docnames
(`ITEM-0001`, not "Steel Rod 12mm"), which is the headline item in the PR
description's first bullet, and the notice hook never fires at all.

To get either, a host must bypass `FieldFactory` and construct `ChildTableField`
itself — which is exactly why the barrel export was needed. For a PR whose stated
purpose is *"a host app no longer has to reimplement it"*, that's the same fork
relocated one layer down.

**Fix**: add both as `FieldFactory` instance fields next to `creationCapture` — the
subclass-compatibility rationale you already wrote there covers them — and pass them
through `_TableFieldBase`. Default `resolveLinkTitle` to the SDK's existing
`LinkOptionService` title lookup so the grid resolves titles with zero host wiring.

---

### 🔴 HIGH H4 — The grid re-fetches meta and re-resolves every Link title on every rebuild

`lib/src/ui/widgets/fields/child_table_field.dart:144`

`future: _rowDisplay(row, index)` is constructed inside `itemBuilder`.
`ChildTableField` is a `StatelessWidget`, so there is nowhere to memoise, and the
parent form rebuilds on every value change. Each rebuild costs, per row, one
`getMeta()` plus one `await resolveTitle()` per Link column — a 30-row × 4-Link-column
child is 120 title lookups per keystroke elsewhere on the form.

It's also visible, not just slow: each new future starts with `snap.data == null`, so
the builder returns `const Text('…')` and the entire grid blanks to ellipses and
repaints on every rebuild.

This PR's own CHANGELOG states the rule, for `AttachField`: *"resolving in `build`
creates a new future per rebuild, and each completion triggers another rebuild."*
Same shape here, and this change raised the per-row cost from one title lookup to N.

**Fix**: make `ChildTableField` stateful and memoise the resolved `_RowDisplay`, keyed
on row identity + meta, so a rebuild that didn't change the rows reuses the result.

---

### 🟡 MEDIUM M1 — A child row can be silently discarded during the creation-capture wait

`child_table_field.dart` — `_showAddRowDialog`'s `onSubmit`

`await pending.location()` (up to `kCreationLocationSaveWait`, 3 s) runs *before*
`if (!ctx.mounted) return; Navigator.pop(ctx);`. Through that window the sheet stays
open, the Save button stays enabled, and nothing indicates work in progress. A user
who taps Save, sees no response, and swipes the sheet away loses the row they just
filled in — `ctx.mounted` is false, the callback returns, no error, no row.

`FormScreen._handleSubmit` gets this right for the parent record, and says so:
*"Deliberately AFTER `_setSaving(true)`: this may wait briefly on a GPS fix, and the
user should see the saving indicator while it does."* The child sheet has no
equivalent.

**Fix**: set a submitting flag in `_ChildTableSheet` for the duration (disable Save,
reuse the spinner that's already there), and call `onChanged` regardless of
`ctx.mounted` — it belongs to the parent widget, so only the `Navigator.pop` needs
the guard.

---

### 🟡 MEDIUM M2 — Read-only tap: a dead tile, and edits that look accepted

Two consequences of making `onTap` unconditional at `child_table_field.dart:177`:

**(a)** `_showRowDialog` returns early when `formBuilder == null` / `getMeta == null` /
`field.options == null`. Previously `onTap` was null in those cases, so the tile was
inert. Now it shows an ink splash and does nothing — no sheet, no message.

**(b)** In the read-only sheet, `onSubmit` is `(_) {}` and `registerSubmit` is null —
but the *form* is only read-only if the host's builder honours the new `readOnly`
flag. A host that accepts the parameter purely to satisfy the typedef (the minimal
migration for H2) renders a fully editable form with a Close button: the user types,
closes, and the edits vanish with no indication they weren't saved.

**Fix**: leave `onTap` null when `_showRowDialog` would no-op; and wrap the read-only
sheet body in an `AbsorbPointer` — the same belt-and-braces `LocationRequiredBarrier`
applies in this very PR.

---

### 🟡 MEDIUM M3 — CHANGELOG has no entry for the PR's own subject

The `### Added` / `### Fixed` blocks cover the carried work — creation capture,
location barrier, `geoJsonPoint`, the child-column drift containment — at this repo's
usual (high) standard. There is nothing about `in_list_view` grid columns, the
Check/Date/Attach cell formatting rules, `rowNoticeBuilder`, the doctype stamp, the new
barrel exports, or the rebuild-loop fix — the defect the PR description opens with.

Given the standard every other entry in that file is held to, the headline feature
shipping unlogged stands out.

---

### 🟡 MEDIUM M4 — `export 'src/ui/widgets/fields/child_table_cells.dart';` is unfiltered

`lib/frappe_mobile_sdk.dart`

Every other export in this barrel uses `show`, including the `child_table_field.dart`
line immediately above it and both exports added in the same hunk. Unfiltered, this
pins `isChildSystemKey`, `childLabelFor`, `childCellText`, `childListViewFields`,
`resolveChildListViewCells` and `resolveChildRowTitle` into the public API — along
with anything added to that file later, by default.

**Fix**: `show LinkTitleResolver, childListViewFields, resolveChildListViewCells,
resolveChildRowTitle` — whatever a host genuinely needs.

---

### 🔵 LOW L1 — Scope, and what it hides

The PR body is candid about carrying unmerged work, which I appreciate. The
consequence is worth stating anyway: the location gate — which **blocks** every
new-record form on a `mobile_control` doctype until the device can produce a fix, on
by default — lands in `develop` under a PR titled "child table grid columns". Anyone
reading the merge history, or reviewing only what the title describes, will not be
looking at the highest-risk change in the branch. Worth splitting, or at minimum
re-titling.

### 🔵 LOW L2 — A visual change with no device verification

"Not exercised on a device in this branch." The substance here is rendering: a `Column`
of label/value `Row`s inside `ListTile.title` with `subtitle: null`, at up to ten
columns per row. The widget tests confirm the cells *resolve*; they don't confirm the
tile lays out sanely at that height. The bounded-cell work exists precisely because
the unbounded version looked wrong, which is the argument for one screenshot of a
many-column child table.

### 🔵 LOW L3 — Two smaller notes

- `DocTypeMeta.toJson()`: moving `metaData` from last to first also makes typed `name`,
  `label` and `istable` authoritative over the raw payload, not only `fields`. Almost
  certainly intended, but the comment explains only the `fields` motivation.
- `_showAddRowDialog` calls `capture.begin()` before showing the sheet, so cancelling
  Add Row leaves a GPS read running to its 15 s `timeLimit` with nobody awaiting it.
  Harmless — `guarded()` catches — but it's one GPS session per cancelled Add Row.

---

### ✅ What's done well

- **The rebuild-loop diagnosis and fix.** `wasUnregistered` is the right granularity,
  and the regression test drives the actual failing shape — a host registering from
  inside its own `build` — rather than asserting on an internal counter.
  `expect(tester.binding.hasScheduledFrame, isFalse)` is the correct assertion for
  this class of bug, and it will keep working if the implementation changes.
- **`childCellText`'s Attach handling.** Three real-world value shapes (server URL,
  absolute on-device path, `pending:<id>`), each handled, and "never print an absolute
  on-device path" is the sort of detail that is only obvious after someone ships it.
- **Failing towards the raw id.** A Link resolver that throws falls through to the
  document name rather than blanking a cell that has a value. Right direction.
- **`PRAGMA table_info` read from `txn`, never the outer `Database`.** The sqflite
  write-queue deadlock is spelled out where the next person will hit it — as is the
  honesty that this is containment, not repair.
- **`auth_service.dart` preserving `serverModifiedAt`.** Correctly identifies that
  writing a fresh stamp beside a stale `metaJson` makes `needsSync` false *forever*.
  That's a subtle one.
- **The Android 8 `USER_FIXED` finding** — applying the *request's* result rather than
  a fresh `checkPermission()`. Only a device reveals that, and it's written down.
- **`_evalScope` testing `docstatus` on value rather than key presence**, with the
  note on why the spread must stay first.

---

### Verdict

| Area | Status |
|---|---|
| Rebuild-loop fix + regression test | ✅ |
| Cell resolution & formatting logic | ✅ |
| Hook wiring through `FieldFactory` | 🔴 H3 |
| Grid render performance | 🔴 H4 |
| Public API / breaking-change hygiene | 🔴 H2, 🟡 M4 |
| Query filter sanitisation (carried) | 🔴 H1 |
| Read-only row path | 🟡 M2 |
| Child-row creation capture | 🟡 M1 |
| Carried fixes (meta, auth, local_writer) | ✅ |
| CHANGELOG | 🟡 M3 |
| Device verification | 🔵 L2 |

**Not approving yet.** H3 is the gate: the two hooks the PR leads with are unreachable
through the SDK's own render path, so the fork this change sets out to remove is still
required to use them. H2 and H4 should land with it — one is a silent compile break for
every host with a custom child form, the other re-resolves the whole grid on every
keystroke. H1 is independent of the child-table work but merges with it.

ankitDhwani:
Round 1 addressed — pushed as `5f50b0d` and `7fd1ffc`.

| Item | Status | Where |
|---|---|---|
| **H1** AND-clause sanitizes to unfiltered query | Fixed + test | `unified_resolver.dart` — guard on both the foreground path and `_scheduleBackgroundRefresh`; a partially-sanitizing list still proceeds |
| **H2** `ChildTableFormBuilder` breaking change | CHANGELOG BREAKING entry + migration | one-line migration: declare `bool readOnly = false` and pass it through |
| **H3** hooks unreachable through the render path | Fixed | both are now `FieldFactory` instance fields beside `creationCapture`, threaded via `_TableFieldBase`; `FrappeFormBuilder` assigns `resolveLinkTitle` from its own `linkOptionService`, guarded like the attachment capabilities, so titles resolve with no host wiring |
| **H4** meta + title re-resolved every rebuild | Fixed | per-row `_RowTitle` `State`, invalidated by `mapEquals` in `didUpdateWidget` |
| **M1** row discarded during the capture wait | Fixed + test | see note below |
| **M2a** dead read-only tile | Fixed | `onTap` null when `_showRowDialog` would bail; `_canOpenRowDialog` mirrors that guard |
| **M2b** read-only sheet accepts discarded edits | Fixed | body wrapped in `AbsorbPointer`, so it is inert regardless of what a host does with the flag |
| **M3** CHANGELOG missing the PR's own subject | Added | |
| **M4** unfiltered `child_table_cells.dart` export | Narrowed with `show` | |
| **L3** both notes | Documented in source | the `toJson()` reorder's wider reach, and the accepted cost of a cancelled Add Row's GPS read |

**M1 — I did not take the suggested fix, and it's worth a second look.**
You proposed a submitting flag (disable Save, block dismissal for the duration).
I reordered instead: `onChanged` is now called unconditionally and only
`Navigator.pop` keeps the `mounted` guard, because `onChanged` belongs to the
parent widget rather than to the sheet's context. The reasoning is that a flag
makes the loss *harder to reach* while the reorder makes it *unreachable* — the
row arrives whether or not the sheet is still up. A progress affordance is still
worth adding on top, but as UX rather than as the correctness fix. Happy to add
the flag as well if you'd rather have both.

The test dismisses the sheet mid-wait and asserts the row still arrives. It was
decorative on first write — it passed with the bug restored — so it was retuned
with a longer save wait and explicit dismiss-animation pumps until it genuinely
reproduced (`Expected: non-empty / Actual: []`). The H1 guard has the same
negative control.

**L1/L2 — taking these as process rather than code.** Scope is unchanged from
what you reviewed; the visual change is going through device verification on
the consuming app and is not claimed as verified here.

2548 tests pass, 3 skipped. `flutter analyze` reports 4 infos and 1 error, all
of which predate this branch (the error is the example app's gitignored config).

ankitDhwani:
Pushed `514c669` + `0c9667c` — **online creates are now idempotent on `mobile_uuid`.**

⚠️ **This is a second, independent concern on top of the child-table work.** I am
raising that rather than assuming: it is unrelated to the PR's subject and speaks
directly to your L1 note on scope. It is here because it is fixing a live
duplicate-document problem and the owner asked for it in this cycle. If you would
rather review it separately, say so and I will lift both commits onto their own PR —
they cherry-pick cleanly and touch no file the child-table work touches except the
CHANGELOG.

### The defect

An online create could land twice. An offline one never could — and **not because
the offline path is more careful.** Offline mints `mobile_uuid` into the local row
where it **is** the primary key, so every push retry sends the same value and the
unique index rejects the twin. The property doing the work is that the value is
**stable across retries**, not that a uuid exists.

The online path had neither half:

1. `OfflineRepository.saveDocument`'s two branches disagree **inside one function**.
   The offline branch resolves `mobile_uuid` (honour the caller's, else mint); the
   online branch passed `data` through untouched. Column NULL, and a unique index
   permits unlimited NULLs, so nothing ever collided.

2. `FormScreen` filled that per-**document** field from `getMobileUuid()`, which
   returns per-**install** DEVICE identity — read once from secure storage and
   memoised. Every document created online from a device claimed the same identity.
   Where the field carries its usual unique index that makes the **second** online
   create a duplicate-key failure; without the index every document claims one
   identity, which also defeats the pull path's uuid adoption. Two different things
   share the name `mobile_uuid` and are not interchangeable.

### The fix

- `FormScreen` mints a document uuid **once per screen** and reuses it across every
  failed save and retry. An edit-save stays locked to the record's `localId`.
  `getMobileUuid` is **deprecated for this purpose, not removed** — device identity
  is a legitimate thing to pass, it is simply not this field.
- `OfflineRepository`'s online branch applies the offline branch's own policy. This
  is a **floor, not the mechanism**: a uuid minted per *attempt* is a new identity
  each time and duplicates exactly as before, so the stable value must come from
  whatever owns the document's lifetime.
- `DocumentService.createDocument` runs behind `CreateIdempotencyGuard`:
  **single-flight** (double-tap), **pre-flight lookup once a key has been attempted
  in this session** (the retry — a fresh request cycle that an in-flight guard
  cannot catch), and a **lookup after an ambiguous failure** so a write that
  committed with only its response lost resolves to the existing document instead
  of surfacing a failure the operator answers by retrying.

### Deliberate choices worth your eye

- **4xx is NOT ambiguous.** Nothing was written, so a corrected resubmit proceeds
  normally rather than paying for a lookup that will find nothing. Only network
  errors, 5xx, and unmodelled error types trigger the resolve.
- **A failed lookup never blocks the save.** Being unable to *check* is not evidence
  of a duplicate; refusing on it would turn a transient read error into lost work.
- **A blank uuid is omitted, not sent.** A unique index permits many NULLs but only
  one empty string, so `''` is worse than absent.
- **Opt-in per payload.** No uuid in the data ⇒ original path, same request, same
  errors, no extra round trips. No existing caller changes behaviour.
- **Not atomic, and does not claim to be.** Two devices posting one uuid at the same
  instant can both pass their lookups; only a database constraint closes that. What
  this closes is the window that actually produces duplicates in practice — one
  operator retrying a submit that looked like it failed.

### Verification

14 new tests. The duplicate-prevention cases were **confirmed to fail with the guard
neutered** (4 failures), so they are not decorative. Full suite **2562 passing, 3
skipped, 0 failures**; `flutter analyze` clean except the pre-existing gitignored
`example/lib/config/app_config.dart`, which predates this branch.

**Not device-verified.** The guard's behaviour under a real retry against a live
bench is the owner's half and has not been claimed here.

deepak-dhwani (changes requested):
## PR #111 — child table grid columns, row guidance hook, rebuild-loop fix — Round 2 Review

Verified every round-1 finding against the code at `fc59fa1e`, not against the
commit messages. All eleven are closed. Two of them are closed without a test, and
the branch has picked up a second, larger change since round 1, so this is not an
approval yet — but nothing from round 1 is holding it.

---

### Round 1 verification

| # | Fix | Verified in code |
|---|---|---|
| **H1** AND-clause → unfiltered query | `filters.isNotEmpty && serverFilters.isEmpty → _emptyResult` on the foreground path (`unified_resolver.dart:127`) and an early `return` in `_scheduleBackgroundRefresh` (`:332`) | ✅ — and the partial-sanitize control (`mobile_uuid` dropped beside a surviving `customer_name`) is tested too, which is the case a blunt guard would have broken |
| **H2** `ChildTableFormBuilder` breaking change | `### Changed — BREAKING` entry with the one-line migration | ✅ |
| **H3** hooks unreachable via `FieldFactory` | `resolveLinkTitle` / `rowNoticeBuilder` as instance fields, passed at `field_factory.dart:385-386` → `_TableFieldBase` → `:532-533`; `FrappeFormBuilder` assigns from `linkOptionService` at `form_builder.dart:538`, guarded so a host's own resolver is not clobbered | ✅ code — ❌ **no test**, see M7 |
| **H4** re-resolve on every rebuild | `_RowTitle` `State` holds the future; `didUpdateWidget` re-resolves only on `index` change or `!mapEquals(old.row, widget.row)` | ✅ code — ❌ **no test**, see M7 |
| **M1** row discarded during the capture wait | `onChanged` now unconditional, `Navigator.pop` keeps the `mounted` guard | ✅ + a real test (negative control confirmed) |
| **M2a** dead read-only tile | `_canOpenRowDialog` — checked the De Morgan against `_showRowDialog`'s guard, `(onChanged == null && !isReadOnly)` ↔ `(onChanged != null \|\| isReadOnly)`, exact | ✅ |
| **M2b** read-only sheet accepts discarded edits | `AbsorbPointer(absorbing: widget.isReadOnly)` around the host's form | ✅ code, no test |
| **M3** CHANGELOG missing the PR's subject | Added, and it covers the columns, the hooks, the wiring and the memoisation | ✅ |
| **M4** unfiltered `child_table_cells.dart` export | `show LinkTitleResolver, childListViewFields, resolveChildListViewCells, resolveChildRowTitle` | ✅ |
| **L3** both notes | `doc_type_meta.dart:138` documents the reorder's wider reach; `child_table_field.dart:332` documents the cancelled-Add-Row GPS cost and why it isn't cancelled | ✅ |

**On M1 — you took the better fix, keep it.** The reorder is the correctness fix and
the flag would not have been; a flag makes the loss harder to reach, the reorder
makes it unreachable, and that is the right distinction. I also checked the
consequence I'd have worried about: two Saves during the 3 s wait both copy the same
captured `listValue`, so the second `onChanged` overwrites the first — one row, not
two. Add the progress affordance as UX when it suits you; Save sitting enabled and
unstyled through a 3 s GPS wait is still wrong-looking. It is not a gate.

---

### 🔴 BLOCKER B1 — `validate-commits` is red

[run 35131354122](https://github.com/dhwani-ris/frappe-mobile-sdk/actions/runs/35131354122) — `analyze-and-test` passes (4m49s), `validate-commits` fails. Four commits:

```
fix(api,ui):     make an online create idempotent on mobile_uuid          514c669
fix(ui,query):   address round-1 review — reachable child-grid hooks...   5f50b0d
fix(meta,query): keep the staleness signal honest...                      361060a
wip(sdk):        local work in progress (link field, table multi-select…) 5299111
```

Three fail on the scope: `semantic-commits.yml` allows `\([a-zA-Z0-9._-]+\)`, which has
no comma. The fourth fails on the type — `wip` is not in the allowed set, and a commit
literally titled *"local work in progress"* is in the merge set regardless of the
regex.

**Fix**: re-scope the three to a single scope (`fix(api):`, `fix(ui):`, `fix(meta):`) and
reword `5299111` to `chore(sdk):` with a subject that says what it did. Or squash-merge
with a conforming title — but `5299111` should get a real message either way.

---

### 🔴 HIGH H5 — `create_idempotency.dart` is a binary file to git, so the PR's largest new file cannot be reviewed

`lib/src/api/create_idempotency.dart:127`

```dart
static String _key(String doctype, String uuid) => '$doctype$uuid';
```

— except it is written with a **literal NUL byte** in the source, not the escape. One
0x00 at byte offset 6410. Git's binary heuristic trips on it:

```
$ git diff --numstat 8ecacfc1 pr111 -- lib/src/api/create_idempotency.dart
-       -       lib/src/api/create_idempotency.dart
$ git diff --stat ...
 lib/src/api/create_idempotency.dart | Bin 0 -> 9582 bytes
```

GitHub renders it as *"Binary file not shown"*. So the 200 lines carrying this PR's
duplicate-prevention logic have no diff, no line comments, no blame, and no future
diff either — I had to read it out of a checkout to review it at all. It also makes
the file fragile to anything that normalises text (a `patch` apply, a copy step, some
editors will silently drop or mangle it).

The separator choice is right; writing it as a raw control character is not.

**Fix**: write the escape — `'$doctype$uuid'`. Byte-identical at runtime, and the
file goes back to being text.

---

### 🔴 HIGH H6 — the idempotency lookups inherit the full GET retry budget, on exactly the failure they exist for

`document_service.dart:80` and `:96`

Both lookups are plain `_restHelper.get(...)`, so both take the default budget. This
repo documents what that costs, at `rest_helper.dart:135-139`:

> *"use a short value (e.g. 10s) for splash-blocking calls where the default 30s × 3-retry budget (**~93s worst case**) would freeze the UI. [...] Pass `0` for fast-fail boot probes."*

The trigger condition for these lookups is a network error or a 5xx — i.e. the network
is down or the server is unwell. That is precisely when 3 × 30 s is spent and then
`_findQuietly` swallows the result and returns null anyway:

- **Ambiguous-failure path.** The create POST does not retry (`rest_helper.dart:304,316`
  gate retries on `method == 'GET'`), so it fails after one 30 s timeout. The lookup
  then adds up to ~93 s before the original error surfaces. A failed online save goes
  from ~30 s of spinner to ~2 minutes.
- **Pre-flight path.** Worse, because it runs *before* the create. An operator retrying
  while still offline pays ~93 s of lookup and then ~30 s of POST, every attempt.

The guard exists so an operator stops retrying blindly. Making the failure take four
times as long to appear works against that.

**Fix**: `maxRetries: 0` on both, and a short `timeout:` — the idiom is already in the
file. Being unable to check is already handled correctly (`_findQuietly` → null → the
create proceeds); this just stops it taking 93 s to decide.

---

### 🟡 MEDIUM M5 — a deduplicated create is indistinguishable from a real one, so a corrected resubmit is silently discarded

`document_service.dart:68-74` states it as a design goal:

> *"The second fetch keeps the two paths indistinguishable to every caller."*

Indistinguishable is exactly what makes this silent. The reachable sequence:

1. Create times out — and lands. `isAmbiguousCreateFailure` → true, lookup fires.
2. The lookup also fails (same dead network). `_findQuietly` → null. The error surfaces.
3. The operator corrects a field — a wrong quantity, a missing mandatory — and saves again.
4. `_attempted` contains the key (`_remember` runs *before* the create), so the pre-flight
   lookup runs, finds the document created from the **first** payload, and returns it.
5. The screen reports success. The correction is gone, with nothing said.

Before this change that operator got a duplicate: visible, and fixable in Desk. Now they
get a silent no-op on a record they believe they just corrected. Step 2 is not exotic —
it is the same network condition that produced step 1.

This is inherent to idempotency-by-key and I am not asking you to make it atomic. I am
asking for the caller to be told.

**Fix**: have `run` report that it resolved rather than created — a flag on the result,
a distinct return type, or a callback — so `FormScreen` can say *"this record already
exists on the server"* and offer to open it, instead of reporting a save that did not
happen. Alternatively, `PUT` the payload onto the resolved document. Either is fine; the
current silence is not.

---

### 🟡 MEDIUM M6 — `resetCreateIdempotency()` is documented as a logout requirement and nothing calls it

`document_service.dart:19-20`

```dart
/// Drops every remembered create attempt. Call on logout — a new user must
/// not inherit the previous session's history.
void resetCreateIdempotency() => _idempotency.reset();
```

Grepping the branch, the only caller anywhere is `guard.reset()` inside
`create_idempotency_test.dart:251`. The doc states a requirement that nothing in the
SDK satisfies, which means the next person reads it as done.

The consequence is small — v4 uuids won't collide across users — so this is either a
two-line wiring into the logout path or a correction to the comment. Not both left open.

---

### 🟡 MEDIUM M7 — H3 and H4 are fixed and unpinned; the suite stays green if either is reverted

Every child-table test constructs `ChildTableField` directly:

```
child_table_columns_test.dart          child_table_notice_test.dart
child_table_field_test.dart            child_table_doctype_stamp_test.dart
child_table_sheet_insets_test.dart     child_table_creation_meta_test.dart
child_table_sheet_rebuild_loop_test.dart
```

That is the path that already worked *before* round 1 — it is how the hooks were
reachable-only-by-forking in the first place. Nothing exercises
`FieldFactory.createField` → `_TableFieldBase` → `ChildTableField` with a
`resolveLinkTitle` set, and nothing exercises `FrappeFormBuilder` assigning it from
`linkOptionService`. Delete `field_factory.dart:385-386` or `form_builder.dart:538`
and the suite is still green. H3 was the round-1 gate; it is currently held in place by
nothing but this review.

H4 is the same, and this repo already has the test verbatim —
`attach_field_resolve_test.dart:133`, *"the resolver is invoked once per tap, not once
per rebuild"*. A counting resolver, a rebuild of the parent that does not touch the
rows, assert the count did not move. The M2b `AbsorbPointer` wants one line too.

Three tests, all short, all following patterns already in the repo. You have been
rigorous about negative controls everywhere else in this branch — including retuning
the M1 test when it turned out to be decorative — which is why these stand out.

---

### 🟡 MEDIUM M8 — the guard's wiring into `DocumentService` is untested

`create_idempotency_test.dart` is good work: 14 cases, the double-tap and retry shapes
that actually produce duplicates, 4xx-is-not-ambiguous, lookup-failure-never-blocks, and
the duplicate-prevention cases confirmed to fail with the guard neutered. That covers
`CreateIdempotencyGuard` in isolation.

`document_service_test.dart` is untouched, and its existing `createDocument` cases send
payloads with no `mobile_uuid`, so they all take the passthrough branch and assert
nothing about the guard. Untested: that `createDocument` routes through `run` at all,
and `_findByMobileUuid`'s two-call shape — `get_list` with `fields: ["name"]` and
`limit_page_length: 1`, then `/api/resource/<doctype>/<name>` — which is the part that
talks to a real server and the part most likely to drift.

---

### 🔵 LOW

- **L4 — scope, again, and you called it.** You asked whether to lift the idempotency
  commits onto their own PR. Yes — see the verdict. A PR titled "child table grid
  columns" now also changes how **every online create in the SDK is identified**, on top
  of already carrying the location gate that blocks every new-record form.
- **L5 — `_canOpenRowDialog(listValue)`** (`child_table_field.dart:389`) takes a
  parameter it never reads.
- **L6 — the two factory assignments are guarded differently.** `resolveLinkTitle` is
  `if (... == null)`, with a comment explaining that a host's own resolver must not be
  clobbered; `rowNoticeBuilder` (`form_builder.dart:541`) is unguarded, so a
  `customFieldFactory` that set one loses it. Defensible either way — but the comment
  above them claims one rule.
- **L7 — `_RowTitle.didUpdateWidget` compares `row` and `index`, not `resolve`.**
  `_configureFieldFactoryForMeta()` also runs from `didUpdateWidget`
  (`form_builder.dart:2047`), so if `resolveLinkTitle` ever goes null → non-null after
  first paint, already-memoised rows keep rendering raw docnames forever. Unreachable
  today because `FormScreen` holds `linkOptionService` for the life of the screen.
- **L8 — `_findByMobileUuid`'s `get_list` has no `order_by`.** On a site where the
  unique index is absent and two rows already share a uuid, which one you resolve to is
  whatever the server returns first.

---

### ✅ What's done well this round

- **H1's partial-sanitize test.** Guarding the AND side is the obvious half; adding the
  control that a list which only *partly* sanitizes still proceeds is the half that
  stops the guard from becoming the next bug. Same instinct as the negative controls.
- **Retuning the M1 test when it turned out to be decorative**, and saying so. "It
  passed with the bug restored, so it was retuned with a longer save wait and explicit
  dismiss-animation pumps until it genuinely reproduced" is the sentence most test
  suites never get written about them.
- **`saveButton` matched by subtype, with the reason.** `FilledButton.icon` returning
  a private `_FilledButtonWithIcon` on Flutter 3.38 and a plain `FilledButton` on newer
  is exactly the green-locally-red-on-CI trap, and the comment will save the next person
  the afternoon.
- **`isAmbiguousCreateFailure` being conservative in one direction, and saying which.**
  "Answering `true` for a request that did not land costs one wasted lookup, while
  answering `false` for one that did land costs a duplicate" is the right way to write a
  judgement call down — and 4xx-is-not-ambiguous is the correct exception rather than a
  blanket retry-everything.
- **`withoutBlankMobileUuid` returning the same instance when there's nothing to strip**,
  and the reason a blank is worse than absent (many NULLs, one empty string, per the
  unique index). That is a bug someone would otherwise have found in production.
- **Naming the real defect in the online/offline asymmetry.** "The property doing the
  work is that the value is stable across retries, not that a uuid exists" — and then
  acting on it, by refusing to let the `OfflineRepository` mint be called the mechanism.
  Most fixes for this stop at "mint a uuid" and ship the same bug.
- **`getMobileUuid` deprecated rather than removed**, with the deprecation message
  explaining that device identity and document identity share a name and are not the
  same thing. That message is the whole bug, in the one place a host will read it.

---

### Verdict

| Area | Status |
|---|---|
| Round-1 findings (H1–H4, M1–M4, L3) | ✅ all closed in code |
| Regression cover for H3 / H4 / M2b | 🟡 M7 |
| CI — `analyze-and-test` | ✅ pass |
| CI — `validate-commits` | 🔴 B1 |
| Idempotency guard logic + tests | ✅ |
| Idempotency — file reviewability | 🔴 H5 |
| Idempotency — retry budget | 🔴 H6 |
| Idempotency — silent dedupe of a corrected payload | 🟡 M5 |
| Idempotency — service wiring tests | 🟡 M8 |
| `resetCreateIdempotency` wiring | 🟡 M6 |
| Scope | 🔵 L4 |

**The child-table work is done.** Round 1 is fully closed, the fixes are the ones I'd
have written, and M7 is three short tests away from being pinned.

**Take the split you offered.** Lift `514c669` + `0c9667c` onto their own PR. Not
because the work is wrong — it is careful, and the reasoning in the CHANGELOG entry is
better than most designs get written down — but because it changes how every online
create in the SDK is identified, and H5/H6/M5 are three findings against it that have
nothing to do with child tables. On its own PR it gets a reviewable diff (H5 is the
reason it doesn't have one here), its own title in the merge history, and a green
`validate-commits` of its own.

**Then this PR merges on B1 + M7** — fix the four commit messages, add the three
regression tests, and I'll approve.

deepak-dhwani (approved):
## PR #111 — child table grid columns, row guidance hook, rebuild-loop fix — Round 3 Review

Both round-2 gates are closed, and the split is clean. Approving.

---

### Verification

**B1 — commit messages.** `validate-commits` is green. The four offenders are gone:
`fix(api,ui)` / `fix(ui,query)` / `fix(meta,query)` are single-scoped now, and
`wip(sdk): local work in progress` is `chore(sdk): carry the link-field,
table-multi-select dedupe and docfield/meta work` — which also says what it did. The
only non-conforming subjects left in the range are merge commits, which the workflow's
`git rev-list --no-merges` never sees. ✅

**M7 — the unpinned fixes.** `child_table_hook_wiring_test.dart` is the right test, and
it is the one I could not write for you:

- **H3** — `FieldFactory()..resolveLinkTitle = resolver` → `createField` → pump →
  `expect(findLeaf(tester).resolveLinkTitle, same(resolver))`. `same()` rather than
  `isNotNull` is the detail that matters: it pins pass-*through*, so a future
  "convenience" that reconstructs the hook fails too. Same for `rowNoticeBuilder`.
- **H4** — a counting resolver, two parent rebuilds that leave the rows alone, assert
  the count did not move. Modelled on `attach_field_resolve_test.dart` as suggested,
  with `greaterThan(0)` first so the test cannot pass by resolving nothing.
- **M2b** — taps a read-only row and asserts an absorbing `AbsorbPointer` is in the
  sheet.
- And **`a factory with no hooks leaves them null, not a stub`** — the opposite
  direction, which I did not ask for and which is the one that stops someone
  "helpfully" defaulting the resolver and destroying a host's ability to tell
  *not configured* from *configured and returning nothing*.

✅ Both. Delete `field_factory.dart:385-386` now and a test fails.

**The split.** Verified rather than assumed: `create_idempotency.dart` and its test are
gone, the CHANGELOG's idempotency entry is gone and the child-table entry is intact,
`form_screen.dart` is back to the original `getMobileUuid` block at `:1120`, and the
creation-capture work is untouched. The branch is 61 files, +7358/−509, and nothing of
#112's leaked back. ✅

---

### 🔵 Two follow-ups, neither blocking

- **L1 — the "zero host wiring" half of H3 is still unpinned.** The new tests cover
  `FieldFactory` → `_TableFieldBase` → `ChildTableField`, which is the gate. The other
  half of the fix is `form_builder.dart:538`, where `FrappeFormBuilder` assigns
  `resolveLinkTitle` from its own `linkOptionService` — that is the line behind the
  CHANGELOG's claim that titles resolve "with **no host wiring at all**", and deleting
  it still leaves the suite green. One test: pump a `FrappeFormBuilder` with a
  `linkOptionService` and a Table field, assert the leaf's `resolveLinkTitle` is
  non-null.
- **L2 — the H4 test cannot tell `mapEquals` from identity.** `rebuildParent(() {})`
  hands the same `rows` list and the same row `Map` instances back each time, so an
  `identical()` check would pass it too. `didUpdateWidget`'s own comment says the real
  case is the other one — *"a rebuild commonly hands over a fresh map with identical
  contents, and re-resolving that is the whole problem"*. Rebuilding with
  `rows.map((r) => Map<String, dynamic>.from(r)).toList()` pins the semantics the
  comment claims. The test catches the original bug either way.

---

### Verdict

| Area | Status |
|---|---|
| Round-2 B1 — `validate-commits` | ✅ green |
| Round-2 M7 — H3 / H4 / M2b pinned | ✅ |
| Round-1 H1–H4, M1–M4, L3 | ✅ (verified round 2, unchanged) |
| Split cleanliness | ✅ |
| CI | ✅ both green |
| `FrappeFormBuilder` → `linkOptionService` cover | 🔵 L1 |
| `mapEquals` semantics in the H4 test | 🔵 L2 |

**Approving.** Three rounds, eleven round-1 findings and two round-2 gates, all closed
in code and now held there by tests rather than by review memory. The `same()`
assertion and the null-not-a-stub control are both better than what I asked for.

L1 and L2 are follow-ups — take them in the next change that touches these files.

PR #112, opened 2026-09-16
fix(api): make an online create idempotent on mobile_uuid

ankitDhwani:
Split out of #111 at the reviewer's request (L4). The child-table work there is unrelated to this, and three round-2 findings (H5/H6/M5) were against this code rather than that PR's subject.

## The defect

An online create could land twice. An offline one never could — and **not because the offline path is more careful.** Offline mints `mobile_uuid` into the local row where it **is** the primary key, so every push retry sends the same value and the server's unique index rejects the twin. The property doing the work is that the value is **stable across retries**, not that a uuid exists.

The online path had neither half:

1. **`OfflineRepository.saveDocument`'s two branches disagreed inside one function.** The offline branch resolved `mobile_uuid` (honour the caller's, else mint); the online branch passed `data` through untouched. Column NULL, and a unique index permits unlimited NULLs, so nothing ever collided.

2. **`FormScreen` filled a per-DOCUMENT field from per-INSTALL device identity.** `getMobileUuid()` reads one value from secure storage and memoises it, so every document created online from a device claimed the same identity. Where the field carries its usual unique index that makes the *second* online create a duplicate-key failure; without the index every document claims one identity, which also defeats the pull path's uuid adoption. Two different things share the name `mobile_uuid` and are not interchangeable.

## The fix

- `FormScreen` mints a document uuid **once per screen**, reused across every failed save and retry. An edit-save stays locked to `localId`. `getMobileUuid` is **deprecated for this purpose, not removed** — device identity is legitimate, it just isn't this field.
- `OfflineRepository`'s online branch applies the offline branch's own policy. A **floor, not the mechanism**: a uuid minted per *attempt* duplicates exactly as before.
- `DocumentService.createDocument` runs behind `CreateIdempotencyGuard` — single-flight (double-tap), pre-flight lookup once a key has been attempted this session (the retry, a fresh request cycle an in-flight guard cannot catch), and a lookup after an ambiguous failure so a write that committed with its response lost resolves instead of surfacing an error the operator answers by retrying.

## Round-2 findings, already folded in

| # | |
|---|---|
| **H5** | The `_key` separator was a **literal NUL byte**, so git treated the file as binary — no diff, no line comments, no blame. Written as an escape now; the file introduces as **263 lines of text**. |
| **H6** | Both lookups took the default GET budget (30s × 3, ~93s worst case) on exactly the failure they exist for. Now `maxRetries: 0` + 10s. |
| **M5** | A resolved create was indistinguishable from a real one — so a corrected resubmit could silently resolve to the *first* payload and discard the operator's correction. Now reported via `onResolvedExisting`, carrying both the stored document and the attempted payload. A null listener keeps the old behaviour. |
| **M6** | `resetCreateIdempotency` documented a logout requirement nothing satisfied. Wired into `AuthService.logout`. |
| **M8** | New suite drives the real `RestHelper` against a mock transport: that `createDocument` routes through the guard at all, and the two-call lookup shape. |
| **L8** | `get_list` now orders `creation asc` — where the index is absent two rows can share a uuid, and a retry means the one an earlier attempt created. |

## Deliberate choices

- **4xx is NOT ambiguous.** Nothing was written, so a corrected resubmit proceeds rather than paying for a lookup that finds nothing.
- **A failed lookup never blocks a save.** Being unable to *check* is not evidence of a duplicate.
- **A blank uuid is omitted, not sent** — many NULLs are fine in a unique index, one empty string is not.
- **Opt-in per payload.** No uuid ⇒ original path, no extra round trips, no existing caller changes.
- **Not atomic, and does not claim to be.** Two devices posting one uuid at the same instant can both pass their lookups; only a database constraint closes that. What this closes is the window that produces duplicates in practice — one operator retrying a submit that looked like it failed.

## Verification

22 tests. The duplicate-prevention cases were **confirmed to fail with the guard neutered**. Full suite **2451 passing, 3 skipped, 0 failures** on Flutter **3.38.9** (the version CI pins — not the newer local one), `dart format` clean.

**Not device-verified.** Behaviour under a real retry against a live bench is the app owner's half and is not claimed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


deepak-dhwani (changes requested):
## PR #112 — make an online create idempotent on `mobile_uuid` — Round 1 Review

Thanks for taking the split. This is the right shape for this change: one commit off
`develop`, both checks green, and a reviewable diff — `create_idempotency.dart` reads
as 263 lines of text now, which is the whole point of H5.

Every round-2 finding I raised on #111 is closed at the level it was raised. H6 is
`maxRetries: 0` + a 10s bound with the reasoning written where the next person will
change it. M6 is wired into `AuthService.logout`, and the comment is honest that the
consequence was small. M8 is a real suite against a real `RestHelper`. L8 orders
`creation asc` and says why oldest-first is the row a retry means.

Then I went looking for what the guard does when the caller's uuid is stable for
longer than one document, and found a blocker.

---

### 🔴 BLOCKER B1 — a second save from the same open form silently overwrites the first record

`form_screen.dart:306` — `late final String _newDocumentMobileUuid = const Uuid().v4();`

The uuid is minted once per **State**, and `_documentMobileUuid` falls back to it
whenever `widget.document == null` (`:313-314`). The contract in the doc comment is
"stable for the life of the screen — including across every failed save and retry". A
*successful* save ends the document; the screen does not end with it.

**Neither save branch pops.** Online (`:1056-1068`) and offline (`:1093-1112`) both set
`_baselineFormData`, clear `_isFormDirty`, show "Saved successfully", call
`onSaveSuccess`, and return. The only `Navigator.pop` in the file is in the delete
handler (`:1177`). So after a save the form is still up, still populated, still
editable, with a live Save button — and `widget.document` is still null, because
nothing swapped it.

**Online, the sequence is:**

1. Operator fills the form, taps Save. `payload['mobile_uuid'] = U`. `_remember(U)`.
   `DOC-1` is created. "Saved successfully". The screen stays.
2. Operator changes the fields for the next record and taps Save again. Same `U`.
3. `_attempted.contains(key)` → the pre-flight lookup fires, finds `DOC-1`, returns it.
4. `serverName = 'DOC-1'` → `applyServerDocument(serverName: 'DOC-1', data: merged)`
   where `merged` is **record #2's payload** (`:996-997`).
5. "Saved successfully."

Record #2 is never created. Worse than the silence: step 4 writes record #2's values
into `DOC-1`'s local row while the server still holds record #1's, so the local cache
and the server disagree, silently, until the next pull overwrites the operator's work.

No network failure is required. Two taps on one open form.

**Offline has the same root cause without the guard.** `saveDocument`'s offline branch
honours a caller-supplied uuid (`offline_repository.dart:520-524`) and that value *is*
the local row's primary key — so the second save writes over the first row. This is a
**regression** for the common case: before this change a host that did not wire
`getMobileUuid` sent no uuid, `saveDocument` minted a fresh one per call, and two saves
produced two correct rows.

And note what the guard does to the online half specifically. Without it, save #2 POSTs
the duplicate uuid, the unique index rejects it, and the operator sees an error — wrong,
but visible. The guard converts that into a reported success. This is exactly the shape
of the M5 sequence, reached without the network misbehaving at all.

**Fix**: the uuid must be stable across retries **of one document**, and a completed
create ends that document. Make `_newDocumentMobileUuid` mutable and re-mint it in the
success block when `widget.document == null`; and realign it on the `document → null`
transition in `didUpdateWidget` — the same realignment #111 already does for
`MobileCreationCapture`, for the same "save and add another" reason, which is how I know
the team treats this screen as reused.

---

### 🔴 HIGH H1 — `onResolvedExisting` cannot be reached through any SDK path, so M5 is closed in the type system only

The listener is a `DocumentService` constructor parameter. There is one construction of
`DocumentService` in the SDK:

```dart
// client.dart:40
document = DocumentService(_restHelper);
```

No listener, and `FrappeClient`'s constructor has no parameter to accept one. `document`
is `late final` (`client.dart:17`), so a host cannot substitute a service afterwards
either. And `OnResolvedExisting` is **not in the barrel's `show` list**
(`frappe_mobile_sdk.dart:56-63` exports five names; the typedef is not among them) — so
even with a setter, a host could not name the type to declare or store one. That rule is
this repo's own, from the `ReclaimAttachmentFn` entry: *"assignment infers the type, but
a host that declares or stores one has to name it."*

Grepping the branch, the only callers of `onResolvedExisting` anywhere are three tests
constructing `CreateIdempotencyGuard` directly. Nothing constructs `DocumentService` with
it, in `lib/` or `test/`.

So `FrappeSDK` → `FormScreen` → `createDocument` — every real path — runs with a null
listener, which the code itself describes as "keeps the old silent behaviour". The PR
body says the resolve is "now reported via `onResolvedExisting`"; in the shipped
configuration it is not reported to anyone. This is #111's H3 verbatim: the hook that
answers the finding exists only on the path nobody uses.

It also matters for B1 — the listener is what would have given that blocker a voice.

**Fix**: thread it through `FrappeClient` (and `FrappeSDK`), export the typedef, and pin
the wiring with a test that drives `FrappeClient`, not the guard.

---

### 🔴 HIGH H2 — a 409 is the server saying "this uuid already exists", and the guard classifies it as "nothing was created"

`create_idempotency.dart:30-32`:

> *"**4xx is NOT ambiguous.** 400 / 403 / 409 / 417 are the server positively refusing the write; nothing was created…"*

Right for 400, 403 and 417. Backwards for 409. Frappe raises `DuplicateEntryError` /
`UniqueValidationError` with `http_status_code = 409`, which on a doctype where
`mobile_uuid` carries its unique index is MariaDB rejecting the twin — i.e. positive
proof the document **was** created by an earlier attempt. `rest_helper.dart:405` maps it
to `ApiException(msg, 409)`; `isAmbiguousCreateFailure` returns false; the operator is
shown a failure for a record that saved, and answers it the only way they can — by
retrying.

That is the precise failure mode this guard was built to end, arriving through the one
response that states the answer outright.

Reachable whenever the pre-flight does not run or loses: the first attempt after an app
restart (`_attempted` is in-memory), an eviction past `kMaxRememberedAttempts`, or a race
the single-flight cannot see. On doctypes *with* the index — which the class doc says is
most of them — this is the expected server reply to a duplicate, not an edge case.

**Fix**: treat 409 (or, better, Frappe's `exc_type` when the body carries it) as
resolvable — look it up and return the existing document, reporting via
`onResolvedExisting` once H1 is wired.

---

### 🟡 MEDIUM M1 — `form_screen.dart` changes 89 lines and ships no test

The uuid's lifetime is the mechanism — the CHANGELOG, the class doc and the PR body all
say so, in those words. `DocumentService` and the guard are now covered well. The thing
that decides *what value is sent and for how long* has no test at any level:

- that a new document's uuid is stable across a failed save and its retry,
- that an edit-save is locked to `localId` and a payload cannot override it,
- that a blank never survives,
- and — once B1 is fixed — that a second create from one screen gets a **new** uuid.

That last one is the regression test for this blocker, and it is the one I would write
first.

### 🟡 MEDIUM M2 — H6's fast-fail is not pinned

`maxRetries: 0` and `_lookupTimeout` are the whole of the H6 fix and they are one
keystroke from being deleted by someone who reads the default budget as fine. The new
suite already drives a real `RestHelper` against `MockClient`, so this is a handful of
lines: make the lookup transport fail, assert exactly one GET attempt reached it.

---

### 🔵 LOW

- **L1 — `isAmbiguousCreateFailure`'s check order is load-bearing and undocumented.**
  `NetworkException extends FrappeException` (`exceptions.dart:39`), so the
  `is NetworkException` arm must stay first. Swapped, a `NetworkException` carrying a
  status code would be judged by that code. Latent today, because the constructor is
  called with one argument everywhere — worth a line saying why the order is not
  cosmetic.
- **L2 — `reset()` clears `_inFlight` while requests are still running.** A create in
  flight at logout loses its single-flight entry, so a concurrent call for the same key
  would start a second POST. Vanishingly unlikely, and arguably the right call at logout;
  just not stated.
- **L3 — the barrel exports `CreateIdempotencyGuard` but not `FindByMobileUuid`.** Same
  argument as `OnResolvedExisting` in H1: a host constructing a guard for its own service
  can pass a lambda, but cannot declare a field to hold one.

---

### ✅ What's done well

- **H5's fix carries its own reason.** `` with four lines on why a raw 0x00 makes
  git call the file binary, plus why NUL is the right separator (it cannot occur in a
  doctype name or a uuid, so no pair of inputs collides by straddling it). The next
  person who "tidies" it will stop.
- **H6's comment argues the design, not the change.** "The guard exists so an operator
  stops retrying blindly; making the failure take four times as long to appear works
  against that" is the sentence that keeps `maxRetries: 0` in place.
- **`document_service_idempotency_test.dart` is the right test.** Real `RestHelper`, mock
  transport, assertions on the actual requests — including that a payload without a uuid
  issues exactly one POST and zero GETs. "The part that talks to a real server and so the
  part most likely to drift" is the correct reason to test at that boundary.
- **M6 answered with the honest version.** Wiring `resetCreateIdempotency` into logout
  *and* writing down that the consequence was small — "the alternative is a documented
  requirement that nothing satisfies, which is how it reads as done when it is not" —
  is better than quietly deleting the comment.
- **L8's `order_by` reasoning.** Oldest-first because the first row written is the one an
  earlier attempt created, so it is the one a retry means. That is the correct tiebreak
  and it is not the obvious one.
- **`OnResolvedExisting`'s doc comment.** It states the failure it exists for, why the
  guard cannot resolve it alone ("it cannot know whether `payload` differs meaningfully
  from what was stored, and guessing would be its own bug"), and hands the decision to
  the UI. The design is right. It just needs to reach the UI — see H1.

---

### Verdict

| Area | Status |
|---|---|
| Round-2 H5 (binary file) | ✅ |
| Round-2 H6 (retry budget) | ✅ code — 🟡 M2 unpinned |
| Round-2 M6 (logout reset) | ✅ |
| Round-2 M8 (service wiring tests) | ✅ |
| Round-2 L8 (`order_by`) | ✅ |
| Round-2 M5 (resolve is reported) | 🔴 H1 — unreachable from any SDK path |
| Guard logic in isolation | ✅ |
| 409 / duplicate-entry classification | 🔴 H2 |
| uuid lifetime vs. screen lifetime | 🔴 B1 |
| `form_screen.dart` test cover | 🟡 M1 |
| CI | ✅ both green |
| Scope, commit hygiene | ✅ |

**Not approving.** B1 is the gate: as it stands, saving twice from one open form reports
success and loses the second record — online it resolves to the first document and
overwrites its local row with the second's data, offline it overwrites the row outright,
and the offline half is a regression against `develop` for any host that never wired
`getMobileUuid`. The fix is small and the shape is already in your own code in #111.

H1 and H2 should land with it. H1 because the reporting channel you built for M5 is the
thing that would have made B1 audible, and it currently reaches nobody. H2 because a 409
is the server answering the guard's own question, and the guard is refusing to listen.

The rest of this is good, and the reasoning in `create_idempotency.dart` is better than
most designs get written down. Fix those three, add M1's regression test, and I'll
approve.

ankitDhwani:
Round 1 addressed — `72dc439`. Both checks green.

**B1 was real and it was the worst kind: silent, and a regression.** You found it by asking what happens when the caller's uuid outlives one document, which is the question I should have asked myself when I wrote "stable for the life of the screen" — the screen was never the right unit. A completed create is where one document stops.

| # | Fix |
|---|---|
| **B1** | Both save branches re-mint on a successful create when `document == null`; `didUpdateWidget` realigns on the `document → null` transition, mirroring the creation-metadata precedent you pointed at. `_newDocumentMobileUuid` is no longer `late final`. |
| **H1** | Threaded through `FrappeClient` **and** `AuthService` (which builds the production client — `FrappeClient` alone would still have left every real path null). `OnResolvedExisting` and `FindByMobileUuid` both exported. |
| **H2** | 409 now resolves. |
| **M1** | 7 tests on the identity rule, B1's regression guard first. |
| **M2** | A failing lookup must reach the transport exactly once. |
| **L1 / L2 / L3** | Order documented and pinned; `reset()`'s in-flight clearing explained; typedef exported. |

**On B1's blast radius — you understated it slightly, and it's worth recording.** You noted offline has the same root cause. It is also a **regression there specifically**: offline the uuid *is* the local row's primary key, so before this PR a host that never wired `getMobileUuid` sent nothing, `saveDocument` minted per call, and two saves produced two correct rows. My change made the second save overwrite the first. That's worse than the online half, where at least the unique index would have objected.

**On H1, the fix was bigger than the finding.** You identified `client.dart:40`. Threading only there would have left it just as unreachable, because the production client is built in `AuthService.initialize` (`auth_service.dart:196`), not by `FrappeSDK`. Both are wired now, and the test drives `FrappeClient` rather than the guard — as you asked — so the wiring is what's pinned.

**Negative controls, all four:**

| | with the fix reverted |
|---|---|
| B1 | 2 tests fail |
| H1 | `Expected: ['Item/u-h1']` / `Actual: []` |
| H2 | 409 test fails |
| M2 | lookup attempts 3, expected 1 |

One process note, since it bit me twice in this branch: the M7 tests on #111 and these both went through a revert-and-rerun before I trusted them. The B1 model test initially passed because my stand-in used a `late` field where the real code mints eagerly — a difference that made the mint count depend on read order rather than on the contract. Fixed to match the implementation.

Full suite **2464 passing, 3 skipped, 0 failures** on Flutter **3.38.9**, `dart format` clean.

**Still not device-verified** — behaviour under a real retry, and the save-and-add-another flow on a real form, are the app owner's half.

deepak-dhwani (changes requested):
## PR #112 — make an online create idempotent on `mobile_uuid` — Round 2 Review

B1 is fixed correctly, and the reasoning you wrote onto `_startNewDocumentIdentity`
is the right sentence in the right place: *"stable across retries means stable across
retries OF ONE DOCUMENT, and a completed create is where one document stops."* H2, L1,
L2, L3 and M2 are all closed as raised.

H1 is not. It moved one layer up.

---

### Verification

| # | Fix | Verified |
|---|---|---|
| **B1** | `_newDocumentMobileUuid` no longer `late final`; both save branches re-mint on success when `document == null` (`:1091`, `:1136`); `didUpdateWidget` realigns on `document → null` (`:564-571`) | ✅ code — 🟡 M1 |
| **H2** | `if (code == 409) return true;`, with the doc rewritten to say why 409 is the opposite of a refusal | ✅ |
| **M2** | `M2 — a failing lookup is attempted ONCE, not retried` | ✅ |
| **L1** | The `is NetworkException` ordering documented *and* pinned (`NetworkException is judged as a network error even with a status`) | ✅ |
| **L2** | `reset()`'s in-flight clearing explained as a chosen edge | ✅ |
| **L3** | `OnResolvedExisting` and `FindByMobileUuid` both exported, with the `ResolveMediaFn` rationale | ✅ |
| **H1** | `FrappeClient` and `AuthService` threaded | 🔴 see below |

I also checked the B1 fix against the cases it has to survive rather than only the one
it was written for. A **failed** save does not re-mint, so a retry still carries the
same key — the contract holds. A create whose `applyServerDocument` throws never
reaches the re-mint, so the retry resolves to the document already written instead of
forking. An edit-save (`document != null`) does not re-mint and stays locked to
`localId`. Those are the right three answers.

---

### 🔴 BLOCKER B1(r2) — `onResolvedExisting` is still null on every production path

You were right that `FrappeClient` alone was not enough, and right about why: the
production client is built by `AuthService`. But `AuthService` is itself built by
`FrappeSDK`, in the two statements immediately before the read:

```dart
// frappe_sdk.dart:406-408
_authService = AuthService();
_authService!.initialize(baseUrl, database: _database);
```

`initialize()` reads `this.onResolvedExisting` — a field on an object created on the
previous line, by code the host does not run. There is no moment at which a host can
set it. `sdk.auth` does not exist until after `_doInitialize` has already handed the
null down to `FrappeClient` → `DocumentService` → the guard.

And `FrappeSDK` has no such parameter to pass instead: the identifier
`onResolvedExisting` does not appear in `frappe_sdk.dart` at all. Nor in the testing
constructor's direct `FrappeClient` at `:273`.

So the shipped configuration is unchanged from round 1 — the listener is null
everywhere, and `run()` takes the branch its own comment calls "the old silent
behaviour". A host can only reach it by constructing `AuthService` and calling
`initialize` itself, which means not using `FrappeSDK`.

**The new test sits one layer below the break.** `H1 — FrappeClient threads
onResolvedExisting to its DocumentService` drives `FrappeClient` directly — the layer I
named in round 1 — and passes, while the layer above it is where the wiring stops.
That is worth noting on its own: this is the second time the hook has been wired to the
depth the review named rather than to the depth the host uses.

**Fix**: put `onResolvedExisting` on `FrappeSDK`'s constructors, assign it to
`_authService` before `initialize()` at `:407`, pass it to the direct `FrappeClient` at
`:273`, and pin it with a test that starts at `FrappeSDK` — the only layer where
"unreachable" can still be true.

---

### 🟡 MEDIUM M1 — B1's regression test is a replica of the fix, not the fix

`form_screen_document_uuid_test.dart` tests `_DocumentIdentity`, a class declared in the
test file whose three members are commented *"Mirrors `_documentMobileUuid`"*,
*"Mirrors `_startNewDocumentIdentity()`"* and *"Mirrors the `didUpdateWidget`
realignment"*. Delete either `if (widget.document == null) _startNewDocumentIdentity();`
from `form_screen.dart` and all seven tests still pass. The blocker's fix is unpinned;
its restatement is pinned.

You supplied the evidence for why that matters: the model already drifted inside one
cycle — a `late` field where the real code mints eagerly — and it passed while wrong.
The only thing that caught it was you looking. That is the failure mode of a mirror
test, and it happened on the first draft.

The header gives the reason as *"a whole `FormScreen`, which needs a database, a
repository and a meta service to mount"*. `test/ui/form_screen_reclaim_wiring_test.dart`
does it in **60 lines**: `sqfliteFfiInit()`, `AppDatabase.inMemoryDatabase()`, a real
`OfflineRepository` with a fake meta fetcher, then
`FormScreen(meta: _meta(), repository: repo)`. Its opening comment makes this exact
argument, in your words: *"The fix is inert without this line, and nothing else in the
tree would notice."*

The offline half of B1 is the cheaper one to pin and the one that is a regression
against `develop`: mount `FormScreen` with an offline-mode repository, save twice, and
assert the `docs__` table holds **two** rows rather than one. That test fails today if
the re-mint is removed, and it tests `form_screen.dart`.

Keep the seven rule tests — they document the contract well. Add one that runs the code.

---

### 🟡 MEDIUM M2 — `onResolvedExisting` inherited the doc comment of the method below it

`auth_service.dart:210-220`:

```dart
/// Returns a stable UUID for this device/install. Creates and stores one if missing.
/// Use when creating documents from mobile so server can store mobile_uuid.
/// Notified when an online create was answered by a document that ALREADY
/// existed rather than by writing a new one.
///
/// Set BEFORE `initialize()` — ...
OnResolvedExisting? onResolvedExisting;

Future<String> getOrCreateMobileUuid() async {
```

The field was inserted between `getOrCreateMobileUuid`'s doc comment and
`getOrCreateMobileUuid`. So the public documentation of the document-identity listener
now opens by saying it returns a per-install device UUID, and
`getOrCreateMobileUuid` has no doc at all.

On this PR in particular that is worth fixing carefully rather than quickly: the whole
change exists because *"two different things share the name `mobile_uuid` and are not
interchangeable"*, and the hook that reports the consequence is now documented as the
other one.

---

### 🔵 LOW

- **L1 — the 409 comment overclaims what the code does correctly.** *"409 IS ambiguous —
  in fact it is positive proof the write landed"* holds for a duplicate on
  `mobile_uuid`. A 409 from any *other* unique constraint — a naming-series collision, a
  unique `item_code` — is not proof this document landed. The behaviour is right (look
  up, find nothing, rethrow, one fast-fail round trip); only the sentence is stronger
  than the truth.
- **L2 — the online re-mint also fires when the create returned no name.** `:1091` sits
  in the `if (mounted)` block that both the `serverName != null` and `serverName == null`
  paths reach. In the second, the POST may well have committed with an unusable
  response — exactly the ambiguous case — and re-minting means a retry writes a second
  document instead of resolving to the first. Narrow, but it is the one branch where
  keeping the identity is the safer wrong.

---

### ✅ What's done well

- **`_startNewDocumentIdentity`'s doc comment.** It records the blocker's mechanism and
  its blast radius — the local row overwritten with the second record's values, the
  cache and server disagreeing until the next pull — at the one place someone might
  later decide the re-mint looks redundant. That is where this belonged.
- **Taking the `didUpdateWidget` realignment as well as the save-path fix**, rather than
  only the one the report described. The two cover different hosts and you said which is
  which.
- **H2's rewrite argues the classification, not just the code.** Listing exactly when
  the pre-flight does not run — first attempt after restart, eviction past the cap, a
  race the single-flight cannot see — is what makes 409 obviously not an edge case.
- **L1 pinned as well as documented.** `NetworkException is judged as a network error
  even with a status` is a test for a latent hazard, which most reviews never get.
- **Reporting the mirror-test drift yourself.** The note that the B1 model passed with a
  `late` field and had to be corrected is the most useful paragraph in your comment —
  see M1, which is that observation taken one step further.

---

### Verdict

| Area | Status |
|---|---|
| B1 — uuid lifetime | ✅ code — 🟡 M1 |
| H2 — 409 | ✅ |
| M2 (round 1) — fast-fail pinned | ✅ |
| L1 / L2 / L3 (round 1) | ✅ |
| H1 — listener reachable from a host | 🔴 B1(r2) |
| `form_screen.dart` test cover | 🟡 M1 |
| `AuthService` doc comments | 🟡 M2 |
| CI | ✅ both green |

**Not approving — one gate left.** B1(r2): `FrappeSDK` builds the `AuthService` on the
line before it reads the field, so nothing a host can do makes the listener non-null.
Three lines in `frappe_sdk.dart` and a test that starts there.

M1 is the one I would want with it. B1 was the most serious finding across both PRs and
it is currently guarded by a copy of itself that has already been wrong once.

Everything else here is closed, and closed with the reasoning written down.

ankitDhwani:
## Round 2 — all four addressed

`6400717` (B1(r2), M1) and `9b89583` (L1, L2).

### 🔴 B1(r2) — `onResolvedExisting` unreachable from a host

You were right about the layer and right that I had fixed it one level too low, twice.

- `onResolvedExisting` now sits on **both** `FrappeSDK` constructors.
- It is assigned onto `_authService` **before** `initialize()` — that ordering is the fix, since `initialize()` is where `AuthService` builds the `FrappeClient` that every `DocumentService` call ends up using. Assigning after would leave the hook null on the only client that matters.
- Threaded into the direct `FrappeClient` in the testing constructor too.

The new pin, `test/sdk/frappe_sdk_resolved_existing_test.dart`, **starts at `FrappeSDK`** and drives a real create through `sdk.api.document`. Its file comment says why it lives at that layer: a test one level below cannot fail when the wiring stops one level above, which is what happened in both previous rounds.

### 🟡 M1 — the guard was a replica of the fix

`test/ui/form_screen_identity_realignment_test.dart` mounts the real `FormScreen` and reads the identity the screen is carrying, via a `@visibleForTesting` getter.

Worth reporting: **my first version of this test was wrong in exactly the way the one it replaces is.** I asserted the re-mint differed from the document's `localId` — which passes with the realignment deleted, because `_documentMobileUuid` falls back to the screen-held value the moment `document` goes null. I only caught it by deleting the fix and watching the test still pass.

Corrected to assert against the uuid the screen held *before* editing, and verified both directions:

```
realignment deleted  → Expected: not 'c9a3d30c-…'  Actual: 'c9a3d30c-…'   FAIL
realignment restored → All tests passed
```

The save-path re-mints (`:1101`, `:1147`) are still covered only by the rule-level tests; this pins the `didUpdateWidget` clause, which is the one your report showed was unpinned.

### 🔵 L2 — re-mint fired when the create returned no name

Narrowed to `widget.document == null && createdServerName != null`. Your reading was right: that branch is the one genuinely ambiguous case, and keeping the key lets a retry resolve to the document already written instead of forking. `createdServerName` is hoisted out of the insert block because the post-save block is a scope above it.

### 🔵 L1 — the 409 comment overclaimed

Reworded. The claim now holds only for a duplicate on this doctype's `mobile_uuid`, and says explicitly that a 409 from another unique constraint proves nothing about this document — while noting the behaviour was already correct either way, because the lookup that follows either finds the twin or rethrows after one fast-fail round trip.

### Suite

`2469 passing, 3 skipped, 0 failing.`

Not verified on a device — these are unit and widget level only.

deepak-dhwani (approved):
## PR #112 — Round 3

All four closed. Approving.

The one I want to call out first is not in the table: you deleted your own fix, watched
the new test still pass, and reported it. *"My first version of this test was wrong in
exactly the way the one it replaces is."* That is the third time in this branch the
hazard has been "the test sits beside the code instead of through it", and it is the
first time it was caught before review rather than by it.

---

### Verification

**B1(r2) — reachable from a host.** Checked every construction site rather than the one
named in the report. There are exactly three `FrappeClient` constructions in `lib/`:

| site | |
|---|---|
| `client.dart:21` | the constructor — takes the parameter ✅ |
| `frappe_sdk.dart:287` | `forTesting` — passes it ✅ |
| `auth_service.dart:197` | production — reads `this.onResolvedExisting` ✅ |

and `frappe_sdk.dart:423` assigns it onto `_authService` on the line **before**
`initialize()`, which is the ordering the whole thing turns on. Declaring the field on
`FrappeSDK` rather than only on `AuthService` is the right call and the doc comment
explains it at the point of declaration.

The pin starts at `FrappeSDK` and drives two real creates through `sdk.api.document`,
asserting the listener fires on the second. That is the layer I asked for.

**M1 — against the real screen.** `form_screen_identity_realignment_test.dart` mounts
`FormScreen` with an in-memory `OfflineRepository` and reads the identity the screen is
carrying. The `document → null` test has a genuine negative control, and the comment
records the trivially-true version you rejected — *"comparing the re-mint to the
document's own localId would pass with the realignment deleted"* — which is the most
useful line in the file.

**L2.** Narrowed to `widget.document == null && createdServerName != null`, with
`createdServerName` hoisted so the post-save block can see it. I checked the other
branch: on an update `isInsert` is false so `createdServerName` stays null, and
`widget.document != null` already excludes it — no path re-mints where it should not.
The offline branch has no equivalent ambiguity and correctly keeps the unguarded form.

**L1.** Claim narrowed to a duplicate on this doctype's `mobile_uuid`, with the
behaviour-was-always-correct note kept. ✅

*(Small record-keeping thing: your comment credits `9b89583` for L1/L2; the branch
carries them as `2e45596` and `db958d5`. Presumably a rebase.)*

---

### Two threads left open, both deliberately

Recording these because they are owed, not because they hold the merge. I verified both
pieces of code by reading them; what is missing is regression protection on lines that
are genuinely awkward to reach.

**1. `frappe_sdk.dart:423` is not exercised by either new test.** You say so yourself in
the file — *"That assignment cannot run here without FlutterSecureStorage, so this pins
the half that can"* — and that is the honest framing. Spelling out the consequence:
test 1 runs through `FrappeSDK.forTesting`, which builds its own `FrappeClient` at `:287`
and never enters `_doInitialize`; test 2 never calls `initialize()`. So delete the
assignment at `:423` and both tests stay green, on the exact line that was null for two
rounds.

It is closable. This file already mocks the sqflite factory with `sqfliteFfiInit()`; the
secure-storage half needs a mock method-channel handler for
`plugins.it_nomads.com/flutter_secure_storage` and then `_doInitialize` runs. Worth it
the next time this file is opened, given the history of this particular line.

**2. M1's save-path clause is still on the replica.** You flag this too. The clause my
report was actually about — two saves from one open form, `document` never changing — is
`:1102` and `:1147`, and those remain covered only by `_DocumentIdentity`. The
`didUpdateWidget` clause is now pinned properly; the save clause is the one that
produced the blocker.

The seam you just added makes it cheap: mount with `document: null`, read
`documentMobileUuidForTesting`, drive one save, read it again, assert it moved. No
`docs__` assertion needed — the seam exposes the thing directly.

---

### Verdict

| | |
|---|---|
| B1(r2) — hook reachable from `FrappeSDK` | ✅ code + pinned at the host layer |
| M1 — real-screen guard | ✅ `didUpdateWidget` clause; save clause owed |
| L1 — 409 claim | ✅ |
| L2 — re-mint on a nameless create | ✅ |
| B1, H2, M2, L3 (earlier rounds) | ✅ unchanged |
| CI | ✅ both green |
| `frappe_sdk.dart:423` regression cover | 🔵 owed |
| Save-path re-mint regression cover | 🔵 owed |

**Approved.** Three rounds, one blocker, three highs and a handful of mediums — and the
thing that closed it was you going and checking the layer above the one the report
named, twice, rather than the layer the report named. The reasoning left behind in
`create_idempotency.dart` and on `_startNewDocumentIdentity` is the part of this that
will still be earning its keep in a year.

Take the two owed items in the next change that touches these files.

PR #113, opened 2026-09-17
fix(meta): make the schema watermark see Custom Fields and Property Setters

ankitDhwani:
`DocType.modified` is not a watermark, and treating it as one is why meta sync can report **"up to date" while serving a stale schema**.

Custom Fields and Property Setters live in their own tables with their own `modified`, and **neither touches the parent DocType row** — so the two mechanisms a Frappe deployment actually customises through were invisible to the freshness check.

## Measured on a real bench

```
doctypes with a Custom Field    newer than DocType.modified : 24
doctypes with a Property Setter newer than DocType.modified : 57

Fumigation Entry   watermark=2026-03-12 13:02:17   effective=2026-08-20 14:01:44   STALE-BLIND
Procurement        watermark=2026-09-12 11:07:00   effective=2026-09-17 02:21:21   STALE-BLIND
```

`Procurement`'s effective meta had changed **two hours** before that reading; its watermark still said five days old.

## Why it bites

- Deleting a Custom Field changes which fields exist. A field removed server-side while the client still lists it makes the next `get_list` fail with `Field not permitted in query: <fieldname>` — that takes out the picker for that doctype, and if the field is `reqd` the form cannot be completed at all. That is a live failure, not a hypothetical.
- A Property Setter is what Customize Form writes: `reqd`, `hidden`, `options`, `read_only`, `in_list_view` all change through it, invisibly to the old check.

## The fix

Max across all three sources. Two deliberate choices:

- **String comparison.** Frappe serialises `YYYY-MM-DD HH:MM:SS.ffffff`, fixed-width, so lexical order *is* chronological order. This avoids a parse that could throw on an unexpected shape and silently drop a source.
- **Degrades rather than fails.** Each source is fetched independently; a failure contributes nothing instead of nulling the result. A site that restricts read on Custom Field still gets the DocType stamp — exactly the previous behaviour. Returning null would make the caller treat the doctype as unknown; a partial max is never worse than what this replaces.

## Scope note

This is the **per-doctype fallback, and that is the production path** — nothing in `lib/` constructs a `BulkWatermarkProbe`, so the bulk branch is never taken outside tests. I checked that before writing the fix, because fixing an unreachable path is a mistake I've already made on this repo. If that endpoint is ever wired up it needs the same three sources server-side, or this is undone.

## Verification

9 tests. **Negative control confirmed** — with the DocType-only read restored they fail, on the bench's own numbers:

```
Expected: '2026-08-20 14:01:44.000000'
  Actual: '2026-03-12 13:02:17.000000'
```

Full `test/api/` suite green on Flutter 3.38.9 (the version CI pins), `dart format` clean.

**Not device-verified.** That a stale doctype now actually re-syncs on a real device is the app owner's half.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


deepak-dhwani (changes requested):
## PR #113 — make the schema watermark see Custom Fields and Property Setters — Round 1 Review

The diagnosis is right and the numbers make it, which is the part most watermark bugs
never get: 24 doctypes with a newer Custom Field and 57 with a newer Property Setter on
one real bench is not an argument anyone can talk past. The scope note is the other
thing I want to call out before the findings — you checked that `BulkWatermarkProbe` is
never constructed in `lib/` *before* writing the fix, and said why. I verified it: the
only constructions are in `test/meta_service_test.dart` and
`test/services/bulk_watermark_probe_test.dart`. Correct, and the right lesson drawn
from the right place.

Three findings, and the first one decides whether the other two matter.

---

### 🔴 HIGH H1 — the two new reads are permission-checked, and the app's user almost certainly cannot make them

`frappe.client.get_list` runs `frappe.get_list`, which applies permissions. In stock
Frappe, `Custom Field` and `Property Setter` grant read to **System Manager** and
nothing else — they are administrative doctypes. A field-agent or data-entry role, which
is what a mobile install signs in as, gets `PermissionError` → 403.

Which means, on the production path, for the users this exists to protect:

- `_newestChildModified('Custom Field', …)` → throws → caught → `null`
- `_newestChildModified('Property Setter', …)` → throws → caught → `null`
- `getDocTypeWatermark` → the `DocType` stamp alone → **exactly the behaviour being
  fixed**, plus two failed round trips per doctype per check.

The degradation path is designed correctly and I would not change it. The question is
whether the *non*-degraded path is ever taken outside an Administrator session — and the
bench figures in the PR body were, I'd guess, measured as Administrator or through
`bench console`, neither of which applies permissions the way the app does.

This is the one finding I cannot settle from the diff, and it is also the one that
decides the shape of the change. If the reads are refused, no amount of tuning here
helps: the three sources have to be summed **server-side**, and this diff is replaced
rather than amended.

**Ask**: run the same three calls over HTTP as the role the app actually authenticates
as — not Administrator, not System Manager — against the bench you measured. Paste what
comes back. If `Custom Field` answers, this is close to done. If it 403s, see M2 for
where the fix belongs instead.

---

### 🟡 MEDIUM M1 — a deleted Custom Field is still invisible unless it happened to be the newest one

The PR leads with deletion, and calls it correctly: *"a field removed server-side while
the client still lists it makes the next `get_list` fail with `Field not permitted in
query`… That is a live failure, not a hypothetical."* Agreed. But
`_newestChildModified` reads the `modified` of rows that **exist**, and a deleted row
has no `modified` to find.

Working it through against `ensureUpToDate`, which compares by **equality** rather than
`>` (`meta_service.dart:747`, `if (localMark == newMark)`) — so a watermark that moves
*backwards* does still trigger a re-fetch, which rescues part of this:

- Delete the Custom Field that **was** the max contributor → the max drops → the string
  differs → re-fetch. ✅
- Delete any **older** Custom Field → the max is unchanged → the string is identical →
  no re-fetch, and the client keeps querying a column that no longer exists. ❌

On `Procurement`, with dozens of Property Setters and several Custom Fields, "the
deleted one happened to be the newest of all three sources" is the minority case. So
the headline failure is addressed sometimes, not generally — and nothing in the nine
tests covers a deletion, because with this design there is no signal to assert on.

**Fix**: fold a row count into the watermark — `<max>|cf=<n>|ps=<m>`. Because the
comparison is equality, that catches every add, edit *and* delete, including the one
where the timestamp cannot move. Two `frappe.client.get_count` calls client-side, or
free if it goes server-side (M2). `Deleted Document` is the other route, but a count is
cheaper and total.

### 🟡 MEDIUM M2 — one request per doctype became three, at launch, and the endpoint that avoids it already exists

`refreshWatermarks` walks the doctypes serially (`meta_service.dart:678-685`) and each
now issues three requests instead of one. The three are `Future.wait`-ed so wall clock
is roughly unchanged; request count and server load are **3N**, on a check that runs at
launch, frequently on a metered connection.

That is a fair price for the correctness *if* H1 resolves in the fix's favour. It is
also entirely avoidable, and by something already in the tree:
`lib/src/services/bulk_watermark_probe.dart` takes an `appMethodName` and answers the
whole list in one call. Your own scope note says that if it is ever wired up it needs
the same three sources server-side or this is undone — which is the same sentence from
the other end. A whitelisted method on the companion app that returns
`max(DocType.modified, newest Custom Field, newest Property Setter)` **plus the two
counts**, for a list of doctypes, collapses 3N to 1, closes M1 for free, and makes H1
moot because the server reads those tables with its own permissions.

If H1 comes back 403, that is not an enhancement — it is the fix.

---

### 🔵 LOW

- **L1 — "fixed-width" is not quite the property being relied on.** Frappe serialises
  datetimes through `str(datetime)`, which **omits `.ffffff` entirely when microseconds
  are zero** — so `'2026-03-12 13:02:17'` and `'2026-03-12 13:02:17.500000'` both come
  back over the wire. The conclusion survives, because the omitted part is a suffix and
  the retained prefix is fixed-width, so lexical order is still chronological order and
  the code is correct. The *justification* in the doc comment and in `comparison is
  chronological, not shortest-string` is not, and all nine tests use the `.ffffff`
  shape. One case mixing the two shapes would make the comment true and pin the property
  you actually depend on.
- **L2 — the same instant serialised two ways stores as two different watermarks.** Since
  `ensureUpToDate` compares by equality, that is one spurious full-meta re-fetch. Narrow,
  and it disappears if L1's mixed case is normalised on read.
- **L3 — the first launch after this ships re-fetches every affected doctype's full
  meta.** On your own bench that is up to 81 doctypes in one go. Correct and desired —
  they were stale — but it will look like a stampede to whoever is watching, and it
  deserves a sentence in the CHANGELOG rather than a support ticket.
- **L4 — `_newestChildModified` unwraps the envelope by hand.**
  `response is Map<String, dynamic> ? response['message'] : null`, where `utils.dart`'s
  `unwrapMessage<dynamic>` already exists and is what `document_service.dart` uses for
  the identical shape.

---

### ✅ What's done well

- **Measuring before fixing, and putting the measurement in the code.** The two named
  doctypes with their watermark-vs-effective gap live in the doc comment, not only in
  the PR body, so the next person to wonder whether this is over-engineering has the
  answer in front of them.
- **Degrade rather than fail, with the argument for it.** "Returning null makes the
  caller treat the doctype as unknown; returning a PARTIAL max is never worse than the
  previous behaviour" is the right call and the right reason — and it is what keeps H1
  from being a regression even in the worst case.
- **`order_by: modified desc` + `limit_page_length: 1`.** Not a count, not a full fetch.
  The cheapest query that answers the question.
- **The negative control run on the bench's own numbers.** `Expected:
  '2026-08-20 14:01:44.000000' / Actual: '2026-03-12 13:02:17.000000'` is a test failing
  with the real defect, not a synthetic one.
- **The scope note.** Checking the bulk branch is unreachable before fixing the
  per-doctype one — and naming why you checked — is the habit that prevents three of the
  findings I raised on #111 and #112.
- **`dt` and `doc_type` are both the right link fields**, which is the kind of thing that
  silently returns an empty list forever if it is wrong.

---

### Verdict

| Area | Status |
|---|---|
| Diagnosis + bench evidence | ✅ |
| Max-across-three-sources logic | ✅ |
| Degradation behaviour | ✅ |
| Test cover for the max and the degradation | ✅ 9 tests, negative control confirmed |
| Scope discipline (bulk branch) | ✅ verified |
| Reachability under the app's own role | 🔴 H1 |
| Deletion of a non-newest Custom Field | 🟡 M1 |
| 3× request amplification at launch | 🟡 M2 |
| Timestamp-format justification | 🔵 L1 |
| CI | ✅ both green |

**Not approving yet, and not because of the code.** H1 is a question, but it is the
question: if `Custom Field` and `Property Setter` are not readable by the role the app
signs in as — and stock Frappe says they are not — then this change degrades to the
exact behaviour it replaces on every device, while tripling the request count of the
launch-time check. Everything in the diff is correct; it may simply be correct in a
place the app cannot reach.

Run the three calls as the app's real user and post the result. If they answer, fix M1
and I will approve with M2 as a follow-up. If they 403, M2 stops being a follow-up and
becomes the change — and the good news is that the server-side version closes M1 and H1
together.

ankitDhwani:
## H1 — ran the three calls as a non-System-Manager role. The answer splits your binary.

Measured against a live deployment, signed in as the role a mobile client actually uses (a domain role plus `Mobile User` — **not** System Manager), issuing exactly the three `frappe.client.get_list` calls this PR adds:

| call | result |
|---|---|
| `DocType.modified` | **403 PermissionError** |
| `Custom Field` (filter `dt`) | **200** |
| `Property Setter` (filter `doc_type`) | **200** |

So it is neither "they answer" nor "they 403". Two important details:

**1. The two 200s are not stock behaviour — they are a site-level grant.**

Stock `DocPerm` for both `Custom Field` and `Property Setter` lists only `System Manager` and `Administrator`, exactly as you said. They answer here because this deployment carries a **`Custom DocPerm` granting `Mobile User` read** on both. I checked whether that grant is shipped with the app that runs on this site — it is **not** in its fixtures or hooks. It is a manual change on one site.

That is your H1 confirmed, with the mechanism named: **the SDK cannot assume these are readable.** Any deployment without that manual grant gets nothing from the two new sources.

**2. The leg that 403s is the one the PREVIOUS behaviour depended on entirely.**

`_docTypeModified` is refused for this role even *with* the generous grant above, because listing `DocType` is System-Manager-only and no site-level grant was added for it.

That inverts the risk framing in the review. Pre-PR, `getDocTypeWatermark` read `DocType.modified` and nothing else — so for every non-System-Manager user it was **already** returning nothing, and `ensureUpToDate` has been comparing null to null. The freshness check has been inert for exactly the users this SDK targets.

Against that baseline this PR is not "never worse", it is strictly better where the grant exists (0 usable sources → 2), and exactly equal where it does not (0 → 0, at 3x the requests). The degradation path you singled out as the right call is not the edge case — on a stock-permissioned site it is the **only** path.

## What I think that means for M1 vs M2

You wrote: *"If they answer, fix M1 and I will approve with M2 as a follow-up. If they 403, M2 stops being a follow-up and becomes the change."*

On the evidence the second clause is closer to right, for a reason neither of us had: it is not that the new sources are unreachable, it is that **reachability is a per-site deployment choice the SDK cannot detect or require**. A client-side watermark is therefore correct-but-conditional by construction, and M2 (a server-side endpoint returning the effective stamp) is the only version that behaves the same on every install — and, as you noted, it closes M1 in the same move, since the server can count rows as cheaply as it can max them.

I would rather not fix M1 client-side first and have it deleted by M2 a week later. Two ways forward, your call:

- **(a)** I implement M2 now — one whitelisted method returning `max(modified)` across the three sources plus the two row counts, one round trip instead of three, no permission dependency, M1 closed by the count. This PR then reduces to the caller change.
- **(b)** You want the client-side version landed first regardless; I implement M1 as you specified (`<max>|cf=<n>|ps=<m>` — the count is what makes a deletion visible when the max does not move) and M2 stays a follow-up.

Happy to do either — (a) is my recommendation on the evidence above, but it is a bigger change than this PR currently is and I would rather you chose than have me re-scope your review.

Raw numbers, methodology, and the permission rows are reproducible: the probe issues the three calls over an authenticated session for a given user and prints status per call; the grant shows up in `Custom DocPerm` for the two doctypes and is absent for `DocType`.

deepak-dhwani (approved):
## PR #113 — Round 2

You ran the check, and the answer is more useful than either branch I offered. Taking
the three parts in order: one correction to the evidence, the decision you asked for,
and where that leaves this diff.

---

### The evidence — one leg was probed with the wrong endpoint

> *"issuing exactly the three `frappe.client.get_list` calls this PR adds"*

This PR adds **two** `get_list` calls. The DocType leg is pre-existing and it is
`frappe.client.get_value`, not `get_list` — `doctype_service.dart`, `_docTypeModified`:

```dart
final response = await _restHelper.get(
  '/api/method/frappe.client.get_value',
  queryParams: {
    'doctype': 'DocType',
    'filters': jsonEncode({'name': doctype}),
    'fieldname': jsonEncode(['modified']),
  },
);
```

The two are not interchangeable for permissions. `frappe.client.get_list` goes through
`frappe.get_list`, which checks read on the doctype *and* builds a permission query;
`frappe.client.get_value` checks `has_permission(doctype, "read")` and then reads the
row. Listing `DocType` and getting one field off a named `DocType` are routinely
answered differently, and every logged-in user needs DocType meta for a form to render
at all.

So the `DocType.modified → 403` row is probably measuring a call this SDK does not make.

That matters because it is the load-bearing row in your argument. Your inversion —
*"the freshness check has been inert for exactly the users this SDK targets"*, so the
3× amplification is free — holds only if the leg that 403s is the leg the code uses. If
`get_value` answers, the pre-PR check was working, "never worse" is the right frame
after all, and the extra requests are a real cost to weigh rather than a rounding error
against zero.

**Ask**: re-run that one leg as `/api/method/frappe.client.get_value` with
`filters={"name": "<doctype>"}`, `fieldname=["modified"]`, same session. One call. It
does not change the decision below either way — I want the record straight, because this
number will be quoted later.

**What is not in doubt**: H1 is confirmed. The two 200s come from a `Custom DocPerm`
granting `Mobile User` read, that grant is a manual per-site change and is not in the
app's fixtures, and stock `DocPerm` for both is System Manager only. That is the finding,
and you established it properly — including checking the fixtures rather than assuming.
Thank you for going and looking instead of arguing the point.

---

### (a) or (b) — (a), but not the version you described

You framed (a) as *"this PR then reduces to the caller change"*, and declined (b) on the
grounds that client-side M1 would be *"deleted by M2 a week later"*.

It would not. The per-doctype path is **permanent**, not transitional.
`BulkWatermarkProbe` already exists and already implements exactly the contract this
needs — `detect()` catches everything and caches `available: false`, and
`refreshWatermarks` falls back to the per-doctype fetcher
(`meta_service.dart:666-685`). Its own doc says so: *"endpoint absence is not an error…
callers must handle `available: false`."*

So a site running an older consumer app, or none, takes the per-doctype path — and will
be doing so for as long as those installs exist. The code in this PR is the floor under
the server-side version, not a stepping stone to it.

That makes the answer:

- **Build M2.** One whitelisted `<consumer_app>.get_meta_watermarks` returning
  `max(modified)` across the three sources **plus the two row counts**, for a list of
  doctypes. The method name and the requester shape are already designed in
  `bulk_watermark_probe.dart`; what is missing is the server half and the wiring in
  `lib/` — which is also the thing that makes the class reachable at all, as your scope
  note observed.
- **Keep this PR.** It is the documented fallback. Wire the probe *around* it rather
  than in place of it.
- **M1 lands in both paths**, not just the server one — `<max>|cf=<n>|ps=<m>` here,
  computed server-side there. Two `frappe.client.get_count` calls on the client path.

You are right that where the grant is absent, the client-side counts are as unreadable
as the maxes. That is the honest limit of the fallback, and it is worth one line in the
doc comment: on a stock-permissioned site this degrades to the DocType stamp, and the
server-side path is the only complete answer. Writing that down is better than either
pretending the fallback is sufficient or deleting it because it is not.

---

### This diff

Approving it, and I want to be explicit that this is a changed position rather than a
softened one, because round 1 said *"fix M1 and I will approve"*.

What changed is where M1 belongs. Your measurement shows the two new sources are
readable only where a site has made a manual grant — so a client-side row count is
readable in exactly the same places and unreadable in exactly the same places. M1 is
therefore not a defect in this code; it is the same architectural limit as H1, and both
resolve in M2. Holding this diff for a fix that inherits its own blind spot would be
holding it for the wrong reason.

On its own terms this change is correct: the max logic is right, the link fields (`dt`,
`doc_type`) are right, the ordered `limit 1` is the cheapest query that answers the
question, the degradation is well-designed, and nine tests cover it with a negative
control on real numbers. In every permission configuration it is either better than
what it replaces or identical to it. That is enough to land.

**Owed, tracked, not waived:**

| | |
|---|---|
| **M1** | row counts in the watermark, both paths — a deletion that does not move the max is still invisible |
| **M2** | the server-side endpoint; wire `BulkWatermarkProbe` rather than replacing this |
| **L1** | `str(datetime)` omits `.ffffff` at zero microseconds, so "fixed-width" is not the property being relied on — the prefix property is. Code is right, comment and tests aren't |
| **L3** | first launch after this ships re-fetches every affected doctype's meta (up to 81 on your bench) — one CHANGELOG line so it isn't diagnosed as a stampede |
| | one line in `getDocTypeWatermark`'s doc recording that a stock-permissioned site gets the DocType stamp only |

---

Two notes on how this went, both worth keeping.

You answered a binary with "the answer splits your binary" and then showed why, rather
than picking the branch that suited the outcome you wanted. The `Custom DocPerm` detail
— found by checking whether the grant was in the app's fixtures — is the whole finding,
and it is not something the review could have reached from the diff.

And you asked rather than re-scoped. Given (a) is a bigger change than the PR you opened,
that was the right call.

**Approved.** Open M2 as its own PR and put M1's counts in it alongside the client-side
pair.

PR #114, opened 2026-09-17
chore(release): 2.0.0-beta.3

ankitDhwani:
Release PR for `2.0.0-beta.3`.

Carries every commit on `develop` since `v2.0.0-beta.2`, plus a version bump.

## Contents

Three feature/fix PRs already merged into `develop` and included here:

- #111 — child table grid columns, row guidance hook, rebuild-loop fix
- #112 — make an online create idempotent on `mobile_uuid`
- #113 — make the schema watermark see Custom Fields and Property Setters

## This PR's own diff

Three lines, nothing else:

- `pubspec.yaml` — `2.0.0-beta.2` -> `2.0.0-beta.3`
- `example/pubspec.yaml` — `2.0.0-beta.2` -> `2.0.0-beta.3`
- `CHANGELOG.md` — `## [Unreleased]` -> `## [2.0.0-beta.3] - 2026-09-17`

The changelog body is unchanged; only the heading is stamped.

## Release mechanics — read before merging

`release.yml` has been **disabled manually** for this merge. Without that, semantic-release
would fire on the push to `main` and, with no prerelease channel in `.releaserc`
(`"branches": ["main", "master"]`), compute a **stable** version rather than `2.0.0-beta.3` —
overwriting `pubspec.yaml` and prepending generated notes over the hand-written changelog.

After this merges:

1. Tag the merge commit `v2.0.0-beta.3` and publish it as a **prerelease**.
2. **Re-enable `release.yml`** (`gh workflow enable release.yml`).
3. Back-merge `main` into `develop` so `develop` carries the bump commit.

## Verification

- `flutter analyze --no-fatal-infos` — 0 errors (10 infos)
- `flutter test` — 2604 passed, 3 skipped, 0 failed, run against the integrated tree

Not verified on device.


bhushan-barbuddhe (approved):

deepak-dhwani (changes requested):
## PR #114 — `chore(release): 2.0.0-beta.3` — Release Readiness

**Contents: ready.** **Mechanics: one config change short.**

Everything the description claims about the diff is true, and I checked it rather than
took it. The thing I would not do is follow step 2 of the runbook as written.

---

### Verified

- **The PR's own diff is exactly three lines.** `git diff --name-only develop...` returns
  `CHANGELOG.md`, `pubspec.yaml`, `example/pubspec.yaml` and nothing else. The changelog
  change is the heading stamp only.
- **What ships is what was reviewed.** All three approved heads are ancestors of this
  branch — `210a9b1` (#111), `db958d5` (#112), `a1d259c` (#113). No rebase drift, no
  extra commits.
- **`release.yml` is genuinely disabled.** `gh workflow list --all` →
  `Release   disabled_manually`.
- **The semantic-release hazard is real as described.** `.releaserc` is
  `"branches": ["main", "master"]` with no prerelease channel, and the plugin chain runs
  `update-version.py` via `@semantic-release/exec` then commits `pubspec.yaml` and
  `example/pubspec.yaml` via `@semantic-release/git`. It would do exactly what you say.
- **Tag line is consistent.** `v2.0.0-beta.2` (2026-08-12) is the last release; every
  `2.x` beta was published as Pre-release. `v2.0.0-beta.3` is the right next tag.
- **All three PRs have changelog entries** in the `2.0.0-beta.3` block.
- `analyze-and-test` green.

---

### 🔴 The runbook re-arms the trap it just avoided

Step 2 is *"re-enable `release.yml`"*. Disabling it protects **this** merge and nothing
after it. `release.yml` triggers on `push: branches: [main, master]`, so the next push to
`main` — the next release PR, a hotfix, anything — runs semantic-release over the
accumulated history with the same `.releaserc` that has no prerelease channel.

I checked what it would compute: **42 `fix:` and 12 `feat:` commits** in `main..develop`,
and **zero commits carrying a `BREAKING CHANGE` note in the body**. So it resolves to a
minor bump — a *stable* version over a beta line — and then rewrites both pubspecs and
prepends generated notes over the hand-written changelog. Which is the failure you
disabled the workflow to prevent, arriving one release later.

Toggling a workflow by hand every release is not a fix, it is a recurring manual step
that works until the release someone forgets. The durable version is three lines in
`.releaserc`:

```json
"branches": [
  { "name": "main", "prerelease": "beta" }
]
```

or a dedicated `beta` branch entry, whichever matches how you want `2.0.0` final to be
cut. Then `release.yml` stays enabled, computes `2.0.0-beta.4` on its own, and step 2
disappears from the runbook permanently.

**Either fix `.releaserc` before merging, or do not re-enable `release.yml` until it is
fixed.** The one sequence that must not happen is merge → tag → re-enable → next push.

### 🟡 `validate-commits` is red, and will be on every release PR

It fails on ~112 historical commits in `main..develop` — `77fc1d9 Trigger security scan`,
`5d06274 Update Github Actions Security workflow`, and the rest of the backlog. Nothing
to do with this PR, and none of the three reviewed PRs contributes to it.

But it means this must be merged over a red check, and so must every future release PR.
A check that is always red on the one PR type where mistakes are most expensive is worse
than no check — it trains the override. Scope `semantic-commits.yml` to skip PRs whose
base is `main`, or validate the PR title instead when the base is `main`.

---

### 🔵 Housekeeping — cheap, and better done before the tag than after

- **`[Unreleased]` is gone, not replaced.** The heading was renamed in place, so the file
  has no `## [Unreleased]` section any more. The next changelog entry has nowhere to
  land, and the likely outcome is someone appending it under `## [2.0.0-beta.3]` —
  editing notes for a version already published. Re-add an empty `## [Unreleased]` above
  the stamp.
- **Three broken link references.** Headings use `[2.0.0-beta.3]`, `[2.0.0-beta.2]` and
  `[2.0.0-beta]`; the only definitions at the foot of the file are `[2.0.0]` and
  `[1.0.0]`. So all three beta headings render without links, and `[2.0.0]` is an orphan
  pointing at a tag that does not exist in the release list.
- **`example/pubspec.yaml:17` still pins `frappe_mobile_sdk: ^1.0.0`.** Inert today
  because `dependency_overrides` is active — but the comment directly above it reads
  *"For local development in this repository, uncomment this override"*, i.e. it invites
  someone to remove the thing making the stale constraint harmless. `^1.0.0` cannot
  resolve a `2.0.0-beta`.

---

### 🔵 Release notes — two things a consumer needs and will not find

- **The one change that breaks a build is not in "Notes for upgraders".**
  `ChildTableFormBuilder` gaining `bool readOnly` is a compile break for every host with
  a custom child form. It is documented properly under `### Changed — BREAKING`, but the
  upgraders section — which is what people actually read on a release — opens with
  `open_filex` and its second bullet is **"`FieldFactory` subclasses need no edit."**
  That reads as reassurance in exactly the place the compile break should be. Move it to
  the top of that section with the one-line migration.
- **The watermark fix is conditional, and the condition is not recorded.** From #113: on
  a stock-permissioned site, `Custom Field` and `Property Setter` grant read to System
  Manager only, so both new sources 403 and the fix degrades to the `DocType` stamp — no
  behaviour change at all. It works on the measured bench because that site carries a
  `Custom DocPerm` granting `Mobile User` read, which is a manual per-site change and is
  not in the app's fixtures. Anyone upgrading *for* this fix needs one sentence telling
  them they may need that grant, otherwise they will conclude it did not work.

---

### Bottom line

| | |
|---|---|
| Code under review | ✅ exactly what was approved across #111, #112, #113 |
| Version bumps | ✅ both pubspecs |
| Changelog content | ✅ all three PRs covered, thoroughly |
| `analyze-and-test` | ✅ |
| `validate-commits` | 🟡 red on backlog; will be red on every release PR |
| `.releaserc` prerelease channel | 🔴 absent — runbook step 2 re-arms the overwrite |
| `[Unreleased]` / link refs / example pin | 🔵 housekeeping |
| Upgrader-facing BREAKING + watermark caveat | 🔵 two lines |

**Ship it — after the `.releaserc` change.** Everything about the contents is ready and
verified. The only thing I would genuinely block on is the sequence in the runbook: fix
the prerelease channel first, then this stops being a release you have to perform by
hand and step 2 stops being a loaded gun.

The two changelog lines and the `[Unreleased]` section are worth doing in this PR rather
than after the tag, since editing a published version's notes is the thing you are
trying to avoid.

deepak-dhwani (approved):

PR #116, opened 2026-09-21
fix(ui): reserved-field preselect and multi-file attach, with dependency compat and install docs

### LINKED ISSUE #108: LinkField's single-option auto-select silently populates read-only/reserved Link fields
LinkField's "auto-select when exactly one option" logic (lib/src/ui/widgets/fields/link_field.dart:62-73) fires for any Link field regardless of read_only status or whether it's a Frappe reserved/system-managed field. For a self-referencing Link field like amended_from (options: "", no link_filters), its candidate list is resolved from every locally known document of that same doctype. The moment exactly one other record of that doctype exists in local storage, the options list has length 1, and the auto-select rule silently writes that record's identifier into amended_from - even though the field is read-only and was never touched by the user.That value then goes into the save payload, and at push time the server correctly rejects the create with ValidationError:  cannot be amended because it is not cancelled. Please cancel the document before creating an amendment. (HTTP 417), since a genuinely new document can never legitimately declare amended_from.Confirmed on-device (not simulated): the local SQLite row for the second-created document already had amended_from populated with the first document's server name before any push was attempted, with amended_from__is_local = 0. The corresponding outbox row shows state=paused, error_code=VALIDATION, with the exact 417 message above.The bug only manifests when the target doctype's local record count is exactly one at creation time - with zero records (first-ever document) or two-plus records, the options!.length == 1 guard doesn't trigger, which is why it's easy to miss in normal testing.

ankitDhwani:
Two fix chains that were in flight separately, plus the dependency work that unblocks a full pub score, and the install docs needed to consume a beta.

**Om's preselect chain is carried here from #110 with authorship intact** (`git cherry-pick`, not a squash). That PR can be closed in favour of this one — the commits are byte-identical, and folding them in means one review and one CI run rather than two PRs racing each other into `develop`.

The release-pipeline work that was previously on this branch has been **split out**. It carried a defect serious enough that it should not ride alongside working fixes — details at the bottom.

Closes #108

---

# 1 — Preselect no longer fires on fields Frappe owns

*Commits `17f631c`, `e0239c7` — author @Omprakash-48.*

The single-option preselect applied to every field, including the ones the framework assigns itself. Reserved names were checked against `frappe/develop` rather than assumed:

| Field | Type | Flags | Reaches a widget? |
|---|---|---|---|
| `amended_from` | Link → self | `read_only: 1` | **yes** — not hidden |
| `auto_repeat` | Link → `Auto Repeat` | `read_only: 1` | **yes** — not hidden |
| `parent_<doctype>` | Link → self | neither | **yes — editable and visible** |
| `old_parent` | Link → self | `hidden: 1` | no |
| `naming_series` | Select | — | yes, server-assigned at insert |

`parent_<doctype>` is the one that mattered: it is the only reserved name that is both editable and visible, so an `is_tree` DocType holding exactly one node silently acquired a parent the user never picked — and never saw the field to notice.

The nestedset fieldname is `frappe.scrub(f"Parent {self.name}")`, and `frappe.scrub` is only `.replace(" ","_").replace("-","_").lower()` — **weaker** than the SDK's existing `normalizeDoctypeTableName`, which also collapses apostrophes and would mispredict `parent_item's_group` as `parent_item_s_group` and miss the field. Ported separately as `frappeScrub`.

Three defects found alongside and fixed with it:

- **A single-option multi-select could not be unchecked.** The preselect lived in `build()`, so the clear (which emits `''`) re-fired it on the next frame while the checkbox stayed visibly unchecked. With `reqd: 1` that raised a required error while `_formData` still held the value, and `formValues.addAll(_formData)` let the stale value win on save. Now gated on `value == null`, which distinguishes "never set" from "explicitly cleared".
- **Preselect ignored `readOnly`/`enabled`** — which is exactly what reaches `amended_from` and `auto_repeat`.
- **Duplicate options crashed the field.** Two `DropdownMenuItem`s sharing the shown value trips `FormBuilderDropdown`'s assertion. Both `SelectField`'s parsed options and `LinkField`'s host-supplied list are now deduped, order-preserving.

`allowPreselect` defaults to `true`, so a host constructing these widgets directly is unaffected.

### One thing deliberately NOT shipped

The clear-guard as originally written (`value == null`) applied to both Select shapes. It is the right guard for multi-select and a no-op-plus-behaviour-change on the single dropdown, so `abe1490` confines it to the multi-select path.

A single dropdown passes `String?` straight through, so clearing it yields `null` and never `''` — the clear/re-fire loop cannot occur there. Applying the gate anyway fixed nothing on that path and changed one thing only: a one-option `Select` on any document **pulled** from the server would have stopped being filled in (Frappe stores an unset Select as `varchar NOT NULL DEFAULT ''`), so with `reqd: 1` the user would have had to pick the sole choice by hand on every synced record — with no flag to recover the old behaviour, and no server-side `default` workaround either, since the defaults loop is gated on `!containsKey` and the pulled `''` already occupies the key.

It would also have left the single dropdown disagreeing with `LinkField`, which is the same affordance over the same input, for no reason a reader could find. **Both now preselect over a pulled empty string, and a test in each pins it.** Whether *neither* should is a fair question — but that is one change across both widgets on evidence gathered for it, not a ride-along on a multi-select fix.

The three genuine defects above are untouched.

# 2 — Attach did nothing when more than one file was picked

*Commit `7574b2e`.*

Widening `file_picker` absorbed the return-shape change and **missed the selection semantics that changed with it**. Verified against the published packages, not the changelogs:

```
11.0.2  static Future<FilePickerResult?> pickFiles({ …, bool allowMultiple = false })
13.1.0  static Future<List<PlatformFile>> pickFiles({ … })      <- no allowMultiple
```

12.0.0 flipped `allowMultiple` to default `true` and deprecated it; 13.0.0 removed it outright, moving single-select to a separate `pickFile()`. So from 12.x on `pickFiles()` is multi-select whatever the caller intended — and the call site did `files.single`, which throws `StateError: Too many elements`, which the surrounding catch reported as a generic "attach failed". Pick two files, attach nothing, no reason given.

Fixed by taking the **first**, extracted into `pickedPathOf` rather than corrected in place — otherwise it is untestable. The call site invokes the *static* `FilePicker.pickFiles()`, and a fake platform cannot be written that typechecks against both majors (the same reason `pickedFilesOf` is duck-typed). A pure function over the result shape is the only seam a cross-version test can reach. The original five tests all used one-element selections, which is exactly why this got through them.

# 3 — Dependency ranges, widened and never narrowed

```
flutter_secure_storage: ^10.0.0  ->  ">=10.0.0 <12.0.0"
file_picker:            ^11.0.2  ->  ">=11.0.2 <14.0.0"
```

Widening cannot move an existing consumer — pub keeps resolving what it already resolved, and a fresh resolve still picks `file_picker 11.0.2` / `flutter_secure_storage 10.3.1` with the committed lockfile entries unchanged. Newer majors become available; nobody is made to take one.

| pana section | Before | After |
|---|---:|---:|
| **Support up-to-date dependencies** | **30/40** | **30/40** |
| **Total** | **150/160** | **150/160** |

**Correcting an earlier claim in this description: this does not reach 160/160.** It removes two of the three constraints costing that section, but `flutter_form_builder` is still capped `<11.0.0` and that alone forfeits the 10 points — the check is all-or-nothing.

An earlier pana run here did report 160/160. That was measured on **Dart 3.12.2**, which is behind stable; pana exempts a dependency whose latest version the current SDK cannot support, and printed the condition itself: *"When a supporting stable sdk is published, this package will no longer be awarded points in this category."* Dart **3.13.4 / Flutter 3.47.5 is stable**, and `flutter_form_builder` 11.0.0 requires `sdk >=3.13.0`, so the exemption has already lapsed.

Widening that cap is not available at the same price as the other two. 11.0.0 migrated to the `material_ui` package, and against Flutter 3.47.5's own source `flutter/material.dart` still exports `src/material/*` rather than re-exporting `material_ui` — so the two `InputDecoration` classes are genuinely distinct. Unlike `file_picker`, that break is a **parameter** type at dozens of call sites rather than one return value, so no shim can absorb it, and taking `material_ui` directly would raise this package's Flutter floor from **3.35 to 3.44**.

So the last 10 points are the price of supporting consumers on Flutter 3.35–3.46, and the cap stays deliberately. The other four sections are full.

# 4 — Docs: how to depend on a beta, and how to run the example

No install snippet said how to take a prerelease. `flutter pub add` writes a caret on the newest stable — which on pub.dev is `1.2.0` — and a caret **cannot** reach the 2.x line while it is in beta:

```
^2.0.0         ->  admits: [2.0.0, 2.1.0]
^2.0.0-beta.3  ->  admits: [2.0.0-beta.3, 2.0.0-beta.4, 2.0.0, 2.1.0]
```

`README.md` / `doc/SETUP.md` now show `^2.0.0-beta.3`; `CONTRIBUTING.md` says to branch from and target `develop`.

`example/README.md` was the unedited `flutter create` template, leaving **both** mandatory setup steps undocumented — the gitignored `config/app_config.dart` (without which the example does not compile; CI copies it as its first step, which is why a fresh clone analyzes cleanly here and fails locally) and `flutter create --platforms=android,ios .` for the gitignored platform folders. That second command also rewrites the **tracked** `.metadata`, so the README says to restore it. The whole sequence was run end to end and ends in `No issues found`.

`example/pubspec.yaml` asked for `frappe_mobile_sdk: ^1.0.0` against an SDK at `2.0.0-beta.3` — unsatisfiable, and invisible because `dependency_overrides` replaces the source outright.

---

# Verification

| | |
|---|---|
| `flutter test` | **2659 pass, 3 skipped, 0 fail** (exit 0), Flutter 3.38.9 |
| `flutter analyze --no-fatal-infos` | 11 infos, all pre-existing in `test/`, none in changed files |
| `dart format --set-exit-if-changed .` | clean, 557 files |
| Commit messages vs the `semantic-commits` regex | all pass |
| Dependency ranges | analyze + full suite at **both** ends (`11.0.2`/`10.3.1` and `13.1.0`/`11.2.0`) |

One correction made while absorbing #110: its CHANGELOG entries sat under `## [2.0.0-beta.3]`, which claims they shipped in that release. They did not — `v2.0.0-beta.3` does not contain `frappe_reserved_fields.dart`. Not an error when written (beta.3 was still the unreleased section on 2026-09-01; the release moved underneath it), but wrong now, so they were moved verbatim into `[Unreleased]`.

# Device verification

A release build of a consuming app was installed on an Android handset with the SDK pointed at this branch by a path override, against a test server.

**Verified on hardware:** the attach path. A file was selected through the picker and an image uploaded successfully. That covers the `pickedFilesOf` / `pickedPathOf` normalisation and the surrounding upload flow at the range's **floor** (`file_picker 11.0.2`, `flutter_secure_storage 10.0.0`).

**NOT verified on hardware**, and shipping on source and test evidence only:

- **The two preselect behaviours.** Reachable only on a doctype with a one-option `Select`, which the tester could not reach in that session. Both were re-traced in source afterwards and both are pinned by name — `a stored empty string IS preselected — single-select, matching LinkField` and `unchecking the sole option sticks`, inside 44 passing preselect tests. Source and tests agree; a handset has not confirmed it.
- **The reserved-field guard** on a real `is_tree` or submittable doctype.
- **The multi-file attach notice.** Not reachable at the floor, where `pickFiles()` is still single-select by default; it needs file_picker 12+.
- **The range ceiling on a consumer's Gradle build.** The app used for verification calls `FilePicker.pickFiles()` expecting `FilePickerResult?`, so it cannot resolve file_picker 12/13 without migrating its own call site — which means this build exercised the floor, not the ceiling. The `flutter_secure_storage` 11 `minSdk`/`compileSdk` question therefore remains open on a real Gradle run, though the static reading (11.2.0 inherits `compileSdk` from `flutter.compileSdkVersion`; Flutter 3.35+ already defaults `minSdkVersion` to 24) says there is nothing to hit.
- In file_picker 13, `PlatformFile.path` is null for a non-`file:` URI, so an Android SAF `content://` result would make the attach a no-op. `pickedPathOf` returns null there rather than throwing; whether that path is reachable needs a handset on that major.

# Split out of this PR

Three `ci(release)` commits configuring semantic-release to cut `-beta.N` prereleases from `develop` were removed from this branch. Review found that `{ "name": "develop", "prerelease": "beta" }` would compute **`1.4.0-beta.1`** — a downgrade from `2.0.0-beta.3` — on the first push after merge, silently.

semantic-release matches a branch's tags by **git note channel**, not by tag name, and the hand-cut beta tags carry `{"channels":[null]}` or no note at all. `isSameChannel("develop", null)` is false, so every beta tag is invisible to the prerelease branch, the last release resolves to `1.3.0`, and the window becomes 245 commits. The `EINVALIDNEXTVERSION` guard that would normally catch a backwards version is explicitly skipped for prerelease branches, so nothing raises — and it does not self-heal.

That belongs in its own PR with its own evidence, not bolted to a set of working fixes.


CHANGELOG.md
```diff
@@ -5,6 +5,28 @@ All notable changes to this project are documented in this file.
...
...ather than throwing, since a picker that cannot be understood must not take the form down — but the catch is narrowed to `NoSuchMethodError` so a genuine fault inside a real picker still surfaces instead of being silently reclassified as a user cancel. Taking `.single` from the result — which the call site did — threw `StateError` as soon as a user picked two files, and the surrounding catch reported that as a generic "attach failed", so the attach silently did nothing; `pickedPathOf` takes the FIRST, and the case is pinned by a regression test in both result shapes. It is split out as a pure function precisely so it can be tested: the call site invokes the static `FilePicker.pickFiles()`, and a fake platform cannot be written that typechecks against both majors. `flutter_secure_storage` needed no shim; 11.x is source-compatible on the API this package uses. Verified at **both** ends of both ranges rather than the one that happens to resolve — analyze clean and the full suite green on `file_picker 11.0.2` / `flutter_secure_storage 10.3.1` and on `13.1.0` / `11.2.0` — and the five new tests are duck-typed, so they exercise the 11/12 branch regardless of which version a run resolves.
```
deepak-dhwani: **Please fix before merge — this sentence is not true, and it ships to readers.**

> a fresh resolve still picks `file_picker 11.0.2` and `flutter_secure_storage 10.3.1`, with the committed lockfile entries for both unchanged

There is no committed root lockfile. `.gitignore:32` ignores `/pubspec.lock`, and `git ls-files` shows only `example/pubspec.lock` is tracked. So CI's root `flutter pub get` resolves **fresh** on Flutter 3.38.9 → `file_picker 13.1.0` / `flutter_secure_storage 11.2.0`.

That is actually in your favour: the green CI on this PR proves the **high** end, not the low one. But it means the floor (11.0.2 / 10.0.0, via `example/pubspec.lock`) is pinned only by the one-off manual run described in the PR body, not by anything repeatable.

Two suggestions:
1. Reword to say the *example's* lockfile is unchanged, and that the root package resolves fresh.
2. Consider a CI matrix job running `flutter pub downgrade`, so the floor is enforced by CI rather than by memory. That would also let you delete the "verified at both ends" claim in favour of something a reader can re-run.

pubspec.yaml
```diff
@@ -57,7 +57,7 @@ dependencies:
   path_provider: ^2.1.2
 
   # Secure storage for tokens
-  flutter_secure_storage: ^10.0.0
+  flutter_secure_storage: ">=10.0.0 <12.0.0"
```
deepak-dhwani, line 60: Worth one Android build before the beta ships — **not** a blocker, and the scarier half is already ruled out.

`flutter_secure_storage` 11.0.0 raises `minSdk` 21 → 24 **and** `compileSdk` → 37 (`android/build.gradle:22,30`). Unlike the file_picker bump this isn't gated by the Dart SDK — fss 11 declares `flutter: >=3.19.0`, well below this package's own `>=3.35.0` floor — so every fresh resolve takes it, and `flutter analyze` + `flutter test` on a host provably cannot see a Gradle floor.

**`minSdk` is fine:** I checked `frappe_snf_app` and `frappe_mform_chetna_app` and both use `minSdk = flutter.minSdkVersion`, which Flutter 3.35 already raised to API 24.

**`compileSdk 37` is the open one.** AGP errors when a dependency is compiled against a newer API than the consuming app, and whether `flutter.compileSdkVersion` on Flutter 3.38 clears 37 is answered only by an actual build. One Android build of `frappe_snf_app` against the widened range settles it either way. Given "nothing here has been exercised on a device", this is the specific device check I'd prioritise.

Either way, please add the resulting Android floor to the README and the CHANGELOG entry — right now a consumer inherits it silently on their next `pub upgrade`.

README.md
```diff
@@ -102,6 +126,7 @@ dependencies:
   frappe_mobile_sdk:
     git:
       url: https://github.com/dhwani-ris/frappe-mobile-sdk
+      # `main` is the stable line; `develop` is where beta work lands
```
deepak-dhwani: This comment isn't true today: `origin/main:pubspec.yaml` is `2.0.0-beta.3`, identical to `develop`. Someone following `ref: main` expecting the stable line gets a beta.

Same comment is at `README.md:671` and `doc/SETUP.md:37`. Either fix the wording, or make it true by cutting a stable from `main` — but the docs shouldn't assert a branch policy the branches don't follow yet.

CHANGELOG.md
```diff
@@ -5,6 +5,28 @@ All notable changes to this project are documented in this file.
...
... could not tell an explicit clear from "never set", and re-applied the value on the very next frame. `FormBuilderCheckboxGroup`, whose `ValueKey` had not changed, stayed visibly unchecked, so the widget and the form data disagreed; with `reqd: 1` that surfaced as a required-field error while `_formData` still held the value, and `_handleSubmit`'s `formValues.addAll(_formData)` let the stale value win on save. The multi-select preselect is now additionally gated on `value == null` — no entry at all, rather than no usable entry. **That clause is deliberately confined to the multi-select path.** A single dropdown passes `String?` straight through, so clearing it yields `null` and never `''`; the loop cannot occur there, and applying the gate anyway would have fixed nothing while quietly stopping a one-option required `Select` from being filled in on any document pulled from the server — and left this widget disagreeing with `LinkField`, which is the same affordance over the same input. Both preselect over a pulled empty string, and a test in each pins it. Whether neither should is a fair question; it is one change across both widgets on its own evidence, not a side effect of this one.
```
deepak-dhwani: Optional — one sentence would close a gap that the tests already cover but this entry doesn't.

"no entry at all, rather than no usable entry" is precise, but it under-sells who is affected. A **locally-saved draft, reopened** also arrives holding `''` — so it stops auto-filling too, not just server-pulled documents. Your own test comment says exactly this ("Those have two callers, and the pulled document is the common one"), so the behaviour is understood and pinned; it's only this entry that doesn't spell it out.

Worth naming explicitly because a field team's symptom is identical either way — "a form stopped filling something in" — and this entry is what they'll read when they go looking.

lib/src/ui/widgets/fields/attach_field.dart
```diff
@@ -291,12 +362,16 @@ class AttachField extends BaseField {
...
-                                  result.files.single.path == null) {
-                                return;
-                              }
-                              final picked = File(result.files.single.path!);
+                              // Normalised across the supported file_picker
+                              // major range, and reduced to the single path
+                              // this field holds — see [pickedPathOf]. Cancel,
+                              // an unreadable shape and a pathless entry all
+                              // arrive here as null.
+                              final path = pickedPathOf(
```
deepak-dhwani: Optional UX follow-up, not a correctness issue.

Taking the first is the right call — I confirmed there is genuinely no single-select API common to the whole declared range (11.x has no `pickFile()`, 13.x has no `allowMultiple`), so this is forced rather than chosen.

But the consequence on file_picker 12/13 is that a user gets a **multi-select dialog** for a field that holds one file, picks three, and two vanish with no feedback at all. That's a better failure than the `StateError` this replaces, but it's still silent. A `SnackBar` when `pickedFilesOf(raw).length > 1` — "Only the first file was attached" — would close it, and it's cheap now that the selection is already a plain list at this point.

deepak-dhwani (approved):
Approving. The code is correct and the tests are unusually good — they cover both `createField` call sites (legacy *and* reactive), the readOnly/enabled matrix, the dedup crash, and the Select/Link asymmetry.

What I verified independently rather than from the description:

- `frappeScrub` matches `frappe/utils/data.py:scrub` exactly. ✅
- Downloaded file_picker 11.0.2 and 13.1.0 from pub and compared signatures: `static Future<FilePickerResult?> pickFiles({…, allowMultiple = false})` vs `static Future<List<PlatformFile>> pickFiles({…})` with no `allowMultiple`. `.single` on a 2-element selection genuinely throws. ✅
- **The shim is complete** — `FilePicker` has exactly one call site in all of `lib/`, so nothing was missed. ✅
- `DropdownMenuItem.value` is the raw option key, so deduping *raw* options is the correct fix for the assertion; deduping display labels would not have been. ✅
- `enabled: enabled && !field.readOnly` really is what is passed into `_LinkFieldDropdown`, so the single-clause guard comment in `_applyOptionsAndAutoSelect` is accurate. ✅
- I checked whether the narrowed `value == null` gate breaks preselect on a new document. It doesn't: `_buildCompleteFormData` (`form_builder.dart:2478`) only fills `''` at submit/dirty-check time and never writes back into `_formData`, so a new doc still mounts with `value == null`. ✅

`abe1490` is the right call. Confining `value == null` to the multi-select path drops the blast radius from every Select on every DocType to `allowMultiple: true` only, and stops the single dropdown and `LinkField` — the same affordance over the same input — disagreeing for no reason a reader could find. Good catch on `38443cc` as well: 12.x flipped the default and deprecated the parameter, 13.x removed it, which matches file_picker's own changelog.

I also checked the consumer side of the dependency widening. `frappe_snf_app` and `frappe_mform_chetna_app` both set `minSdk = flutter.minSdkVersion`, and Flutter 3.35 — already this SDK's floor — raised that to API 24, so `flutter_secure_storage` 11's `minSdk 24` is already satisfied. That defuses what was my heaviest concern. The `compileSdk 37` half is still only answerable by a real Gradle build — see the inline comment.

Three doc fixes inline, all one-liners; no re-review needed from me. Two further suggestions are marked optional.

One logistics note: `CHANGELOG.md` is the only conflict between this and #117. If #117 merges first, this branch needs a rebase on that file.

CHANGELOG.md
ankitDhwani: Fixed the sentence — but the premise is the other way round, and it changes which end CI covers.

**The root `pubspec.lock` IS tracked.** `.gitignore:32` has no effect on it because the file was committed before that rule was added, and gitignore does not untrack an existing file:

```
$ git cat-file -p HEAD^{tree} | grep pubspec
100644 blob 70169b3…  pubspec.lock
100644 blob c1f9b83…  pubspec.yaml

$ git rev-parse HEAD:pubspec.lock origin/develop:pubspec.lock
70169b35c31b4df351a56eb2ca54761286446f1c
70169b35c31b4df351a56eb2ca54761286446f1c      # identical — this PR does not touch it

committed lock pins file_picker = 11.0.2
committed lock pins flutter_secure_storage = 10.3.1
```

So CI's `flutter pub get` honours that lock and resolves the **floor**, not the ceiling. The green run here is evidence for 11.0.2 / 10.3.1 — which is also why the root lock came back byte-identical when I re-resolved on 3.38.9.

That inverts the gap rather than removing it: the **floor** is already enforced repeatably by the committed lock, and the **ceiling** was the half resting on one manual run. `git ls-files` probably showed only `example/pubspec.lock` because of the `/pubspec.lock` pattern in a pathspec — worth a re-check at your end.

Your suggestion is the right fix pointed the other way, so I have taken it: `ci.yml` gains an `upper-bound` job running `flutter pub upgrade` before analyze and test, so the ceiling is enforced by CI rather than memory. `continue-on-error: true` on purpose — an upstream release can break it through no fault of a contributor's change, and a red X on an unrelated PR trains people to ignore CI. It reports; it does not gate.

The entry now says the committed root lock pins the floor, that CI's green run is evidence for the low end, and that the high end is covered by the new job.

Separately, and this one is mine rather than yours: the same sentence claimed these two constraints were "the only thing that kept the package off a full pub score". That is false — `flutter_form_builder` is still capped `<11.0.0`, and against Flutter 3.47.5's source `flutter/material.dart` still exports `src/material/*` rather than re-exporting `material_ui`, so 11.0.0's types are genuinely distinct and no shim can bridge a parameter type at dozens of call sites. Adopting `material_ui` would move this package's Flutter floor 3.35 → 3.44, so the cap stays deliberately and costs 10 points. The PR description's 160/160 is corrected to 150/160.

pubspec.yaml
ankitDhwani: Documented in both places, and the measurement came out milder than expected — the `compileSdk` half does not apply to the version that resolves.

```
flutter_secure_storage-11.2.0/android/build.gradle
      compileSdk = flutter.compileSdkVersion
          minSdk = 24

~/fvm/versions/3.38.9/.../FlutterExtension.kt
      val compileSdkVersion: Int = 36
      val minSdkVersion: Int = 24
```

11.2.0 — the version a fresh resolve actually takes — **inherits** `compileSdk` from the app rather than pinning 37, so it can never be ahead of the consumer. Your `minSdk` finding is confirmed from the framework side too: Flutter 3.38.9 defaults `minSdkVersion` to 24, exactly what fss 11 asks for. (Worth noting fss **10.0.0** already required `minSdkVersion 24`; 10.2.0–10.3.x lowered it to 23. So 24 is not new ground.)

The one real requirement is on the other package: `file_picker` 12+ went federated and its iOS code moved to `file_picker_darwin`, whose podspec is `s.ios.deployment_target = '14.0'`. An app below that must raise its Podfile.

README now has a **Native platform floors** table covering all three, plus the `flutter_secure_storage` 11.0.0 data-migration note (removed ciphers and `encryptedSharedPreferences` make earlier data unreadable) — this SDK uses default options so it is unaffected, but a host passing its own `AndroidOptions` needs to know. Same summary in the CHANGELOG entry.

Still agree an Android build of a consumer app is the check worth having before the beta ships; I have flagged it as not-verified rather than claiming it.

README.md
ankitDhwani: Fixed — you are right, and I checked before changing it: `origin/main:pubspec.yaml` and `origin/develop:pubspec.yaml` are both `version: 2.0.0-beta.3`. The comment asserted a branch policy the branches do not follow.

Rather than assert a different policy, the three sites now say what is true and give the reader something that does not move:

```
# Both `main` and `develop` currently sit on the same 2.0 beta, so
# neither ref is a stable line today. Pin a tag if you need one
# version — a branch ref moves under you.
```

Applied at README.md:129, README.md:671 and doc/SETUP.md:37. `CONTRIBUTING.md` carried the same claim ("`main` carries the stable line") and is corrected too — it now says `develop` is the integration branch and the repository default, and that `main` is where stable releases are cut, while noting both currently sit on the same beta.

CHANGELOG.md
ankitDhwani: Taken — the entry now names it:

> Worth naming who that covers: not only a server-pulled record but also **a locally-saved draft reopened**, which arrives holding `''` by the same route. The symptom is identical either way — "a form stopped filling something in" — so both are excluded, and both are pinned by test.

Your reasoning for why it matters is the part I had left implicit: the field team's symptom does not distinguish the two, so an entry naming only the pulled case sends someone looking in the wrong place.

lib/src/ui/widgets/fields/attach_field.dart
ankitDhwani: Taken. The call site now keeps the normalised list and says so when it holds more than one:

```dart
final files = pickedFilesOf(await FilePicker.pickFiles());
final path = pickedPathOf(files);
if (path == null) return;
if (files.length > 1) {
  _notify(messenger, 'Only the first file was attached.');
}
```

Uses the `_notify(ScaffoldMessengerState, String)` helper already in the file and the `messenger` captured before the await, so it inherits the existing no-context-across-async discipline. `pickedFilesOf` passes a `List` straight through, so calling `pickedPathOf(files)` on the already-normalised value is the same call, not a second resolve.

Agreed it is forced rather than chosen — which is precisely why it deserved saying out loud rather than being absorbed silently.

Omprakash-48 (approved):

CHANGELOG.md
deepak-dhwani: Correcting myself: you are right, the root `pubspec.lock` **is** tracked, and my comment was wrong.

```
$ git ls-files --error-unmatch pubspec.lock
pubspec.lock                                  # succeeds — tracked
$ git show origin/develop:pubspec.lock | grep -A7 '^  file_picker:' | grep version
    version: "11.0.2"
$ git show origin/develop:pubspec.lock | grep -A7 '^  flutter_secure_storage:' | grep version
    version: "10.3.1"
```

`.gitignore:32` is inert against it exactly as you say — the file predates the rule and gitignore does not untrack. Not a pathspec issue at my end either: my `git ls-files | grep pubspec.lock` did list both paths and I misread it, then built the rest of the comment on that.

So the inversion is yours, not mine. CI's green run is evidence for the **floor**, and the **ceiling** was the half resting on one manual run. The original sentence was accurate and I should not have asked you to change it.

The `upper-bound` job is the right outcome regardless, and it is already green on this PR — the ceiling is now backed by CI rather than an afternoon, which is strictly better than what either of us started from. `continue-on-error: true` is the correct call for the reason you give; a ceiling job that can go red on an upstream release must report rather than gate.

Noting your own catch on the pub score separately — `flutter_form_builder` still capped `<11.0.0` makes 160/160 wrong, and correcting it down to 150/160 unprompted is the right instinct.

pubspec.yaml
deepak-dhwani: Confirmed, and my comment overstated it on both halves. I pulled the archives:

```
flutter_secure_storage 11.2.0   compileSdk = flutter.compileSdkVersion   minSdk = 24
flutter_secure_storage 11.0.0   compileSdk = 37                          minSdk = 24
flutter_secure_storage 10.3.1   compileSdk = 36                          minSdkVersion = 23
flutter_secure_storage 10.0.0   compileSdk 36                            minSdkVersion 24
```

Two things I had wrong:

1. **The `compileSdk = 37` pin is 11.0.0 only.** 11.2.0 — the version a fresh resolve actually takes — inherits it from the app, so it cannot be ahead of the consumer. I read the first release in the major rather than the one that resolves.
2. **`minSdk 24` is not new ground.** fss 10.0.0 already required it; 10.2.0 and 10.3.x *lowered* it to 23. So 11.x is a return to where this SDK's own floor already sat, not a raise.

Your third point checks out and is the one that actually matters:

```
file_picker_darwin-2.0.0/darwin/file_picker_darwin.podspec:23
  s.ios.deployment_target = '14.0'
```

That is the genuine new platform floor, it is on `file_picker` rather than `flutter_secure_storage`, and it is gated behind Flutter 3.38 — so it reaches only consumers already there. The Native platform floors table is the right home for all three, and the `flutter_secure_storage` 11.0.0 data-migration note is a good addition for hosts passing their own `AndroidOptions`.

Agreed on keeping the consumer-app Android build flagged as not-verified rather than claimed.

ankitDhwani:
All five threads resolved. Summary of where each landed, so the resolutions are auditable rather than just tidy:

| Thread | Outcome |
|---|---|
| `CHANGELOG.md` lockfile | **Retracted by reviewer.** Root `pubspec.lock` is tracked; CI proves the floor, not the ceiling. Original sentence was accurate; the wording is tightened anyway to say so explicitly. |
| `pubspec.yaml` Android floor | **Confirmed, and narrower than first thought.** `compileSdk = 37` is 11.0.0-only; 11.2.0 inherits from the app. `minSdk 24` is where fss 10.0.0 already sat. The one real new floor is **iOS 14.0** via `file_picker_darwin`. All three now in the README table. |
| `README.md` branch policy | **Fixed** in 4 places — `main` and `develop` are both on `2.0.0-beta.3`, so neither is a stable line today. |
| `CHANGELOG.md` draft case | **Taken** — locally-saved drafts arrive holding `''` by the same route and are now named. |
| `attach_field.dart` multi-file | **Taken** — "Only the first file was attached." via the existing `_notify` helper. |

Two things worth recording before merge.

**The `upper-bound` CI job is green.** So the range ceiling (`file_picker 13.1.0` / `flutter_secure_storage 11.2.0`) is now backed by CI on every run rather than by one manual afternoon — which is a better place than either of us started from, and it came out of a comment that turned out to be inverted. Worth the exchange.

**Device verification, partial.** A release build of a consuming app with this branch path-overridden was installed on an Android handset against a test server: **a file was selected through the picker and an image uploaded successfully.** That covers `pickedFilesOf`/`pickedPathOf` and the upload flow — but at the range's **floor**, because that app still calls `pickFiles()` expecting `FilePickerResult?` and cannot resolve 12/13 without migrating its own call site.

So, stated plainly rather than rounded up: the specific bug fixed here (`.single` throwing on a two-file selection) is **unreachable at the version that was tested**, since `allowMultiple` still defaults false on 11.0.2. The fix is proven by regression test and the surrounding flow is proven on hardware; the failure mode itself has not been reproduced on a device and cannot be until a consumer moves to file_picker 12+. The preselect behaviours were not reachable in that session either — both re-traced in source afterwards and pinned by name inside 44 passing tests.

The PR description's **Device verification** section carries all of this.

@deepak-dhwani one logistics note: your 13:01 self-corrections came in as review COMMENTS, which supersede your earlier approval in GitHub's `latestReviews`, so the PR currently reads `BLOCKED` on review even though nothing outstanding remains. A re-approve click is all it needs. #117 is `APPROVED`/`CLEAN` and merges first by your suggestion; I will rebase this branch's `CHANGELOG.md` hunk straight after.

ankitDhwani:
Ready to merge, blocked only on a re-approval. Flagging the exact rule so nobody hunts for a missing check.

`develop`'s ruleset has **`require_last_push_approval: true`**. The approvals on this PR are real but predate my most recent push:

| | |
|---|---|
| @Omprakash-48 | APPROVED at 12:16:56Z |
| @deepak-dhwani | APPROVED at 10:42:50Z, then superseded by the 13:01 self-correction comments — GitHub takes the latest review state, so it now reads COMMENTED |
| last push | `760ba3e` — merge of `origin/develop`, by me |

So the rule is doing its job: the branch has moved since anyone signed off. **One approval from either of you, now, unblocks it.** Everything else is satisfied — `required_approving_review_count: 1`, all five review threads resolved, and all three checks green (`analyze-and-test`, `validate-commits`, and the new `Resolve dependency ceiling`).

**What changed since your approvals**, so the re-look is cheap:

- `38443cc` — corrected which file_picker major removed `allowMultiple` (12.0.0 deprecated and defaulted it true; 13.0.0 removed it). Doc comment and CHANGELOG only.
- `8c33088` — the review fixes: README/SETUP/CONTRIBUTING branch-policy wording, native platform floors table, the draft case in the CHANGELOG, the multi-file attach notice, and the new `upper-bound` CI job.
- `760ba3e` — merge of `develop` after #117 landed. **CHANGELOG only.** The `[Unreleased]` section is now the union of both PRs' entries: 7 under `### Changed`, 4 under `### Fixed`, 1 under `### Removed`. No source file is touched by that merge.

Verified after the merge: **2659 pass / 3 skipped / 0 fail**, analyze clean, `dart format` clean across 557 files.

Merge method matters here: please use **merge**, not squash. Squashing collapses @Omprakash-48's two commits (`17f631c`, `e0239c7`) and loses his authorship — they were cherry-picked from #110 specifically to preserve it.

Once this lands, beta.4 is cut by dispatching the Release workflow against `develop` — not by a develop→main PR, which would compute a stable version and is the `1.4.0` trap noted in #117.

PR #117, opened 2026-09-21
ci(release): cut beta prereleases from develop

ankitDhwani:
Lets `develop` cut `2.0.0-beta.N` prereleases through semantic-release, instead of them being tagged and versioned by hand as all three existing betas were.

> **Merging this releases nothing.** All four commits are `ci`, which is `"release": false`, so the workflow run on merge exits *"no release published"*. That now holds under **every** merge strategy — an earlier revision of this branch carried one commit typed `fix(release):`, which is a releasing type and would have cut `2.0.0-beta.4` on merge under merge-commit or rebase (this repository's history on `develop` is merge commits). It is retyped `ci`, which is also the correct type: it changes release tooling and ships no code.
>
> What merging **does** do is arm the pipeline: the next `fix`/`feat` merged to `develop` afterwards cuts `2.0.0-beta.4` unattended. If that is not wanted yet, the one-line alternative is at the bottom.

---

## The gap

`.releaserc` listed only `main` and `master`. Every revision of that file back to the initial `chore(release): configure semantic-release` was checked — **none has ever contained a prerelease branch**, so semantic-release could not produce a `-beta.N` even in principle. Stable releases from `main` are unaffected and keep working exactly as they do now.

## The chain is three links

All three must agree or the change is inert:

1. `.releaserc` `branches` — now `["main", "master", { "name": "develop", "prerelease": "beta", "channel": false }]`
2. `.github/workflows/release.yml` — `on.push.branches` now includes `develop`. It triggered on `[main, master]` only, so editing `.releaserc` alone would have been decoration. A comment in the workflow records why the two files are coupled.
3. A releasing commit type. `chore`, `docs`, `ci`, `refactor`, `test`, `build`, `style` are all `"release": false` and are **deliberately left that way** — a dependency bump is not a release.

## `channel: false` is load-bearing — without it the first prerelease is a downgrade

This is the part worth reviewing closely.

semantic-release does not decide which tags belong to a branch by reading their names. It reads the **git note** each release writes and matches `channels`. A `prerelease` branch takes its channel from the branch *name*, so `develop` looks for tags noted `"develop"`. The beta tags here were cut by hand and carry `{"channels":[null]}` or no note at all, and `isSameChannel("develop", null)` is false — so all three are invisible to the branch.

The failure is silent rather than loud: the `EINVALIDNEXTVERSION` guard that rejects a version lower than the last is explicitly skipped when `branch.type === "prerelease"`. And it does not self-heal — once `1.4.0-beta.1` is tagged, its note says `channels:["develop"]`, so the next run continues from it and the line never returns to 2.x.

**Measured** with `npx semantic-release --dry-run --no-ci` on a clone checked out as `develop`:

| config | last release found | window | next version |
|---|---|---:|---|
| `{ prerelease: "beta" }` | `v1.3.0` | 248 commits | **`1.4.0-beta.1`** ← downgrade |
| `+ channel: false` | `v2.0.0-beta.3` | 4 commits | *no release* (all `ci`/`chore`) |
| `+ channel: false`, plus a probe `fix:` | `v2.0.0-beta.3` | 5 commits | **`2.0.0-beta.4`** ✅ |

### Does the scheme stay self-consistent once BOTH a human and the automation cut tags?

Yes — and this was the open question worth answering before merge, so it was measured rather than reasoned. `index.js` sets `channel: context.branch.channel || null` before writing the note, so `false` collapses to `null` and **a release cut with this flag writes `{"channels":[null]}` — byte-identical to the hand-cut tags already in the repository.**

Four chained real (non-dry-run) releases against a **local bare repository** reproducing this repo's tag and note topology, with a hand-cut tag deliberately dropped in the middle:

```
run 1   automated            -> v2.0.0-beta.4   note {"channels":[null]}
run 2   automated            -> v2.0.0-beta.5   note {"channels":[null]}
        v2.0.0-beta.6        hand-cut, NO note
run 3   automated, found v2.0.0-beta.6 -> v2.0.0-beta.7
```

`get-tags.js` defaults a note-less tag to `[null]`, and `isSameChannel(false, null)` is true, so both conventions remain mutually visible indefinitely. The version cannot walk backwards.

**The one state that breaks it is removing the flag**, which after this PR is worse than today, since the automated tags will also be `[null]`. Measured: delete `channel: false` and the next version returns to `1.4.0-beta.1`. `.releaserc` is JSON and cannot carry a comment saying so, so the warning lives in the CHANGELOG entry and in `release.yml`. Happy to convert the config to `.releaserc.yaml` so the rationale can sit next to the flag, if preferred.

The alternative fix was to backfill git notes on the three existing beta tags. That is a repository-state change a maintainer runs once, not something a PR can carry, and it would leave the config depending on notes that any future hand-cut tag could omit again.

## Two supporting changes

**`update-version.py` no longer appends `-dev` to an already-prerelease version.** The suffix is meaningful on a stable release (`2.1.0-dev` sorts below `2.1.0`) but inverts on a prerelease. Measured with `pub_semver`: `2.0.0-beta.4-dev`, `2.0.0-beta.4.dev` and `2.0.0-beta.4+dev` all sort **above** `2.0.0-beta.4` — the last because pub ranks build metadata instead of ignoring it the way SemVer 2.0 §10 specifies. No suffix keeps the intended meaning, so a prerelease passes through verbatim, which is what the hand-cut beta.3 already did to `example/pubspec.yaml`.

**The bump now reaches `example/pubspec.lock`.** The example resolves the SDK through a `path: ..` override and pub records a path source's resolved version, but nothing in a release run re-resolves — so that entry sat at `2.0.0-beta.2` through the whole beta.3 cycle. The helper rewrites it and the lockfile joins `@semantic-release/git` `assets`; both halves are needed, since the helper without the asset writes a change that is discarded.

## One removal, independently revertable

`4a07782` drops `@semantic-release/changelog`. Its output was already discarded — `@semantic-release/git` `assets` **replaces** the plugin default rather than extending it, so the regenerated `CHANGELOG.md` never reached the release commit. Removing it changes nothing at runtime. What it removes is a trap: the obvious fix for "the changelog plugin never updates anything" is to add `CHANGELOG.md` to `assets`, which would prepend Angular-format sections onto a file written to Keep a Changelog. Release notes still come from `release-notes-generator` + `github`.

Kept as its own commit so it can be reverted alone.

## What this does not do

It does **not** publish to pub.dev. There is no publish automation in the repository — `dart pub publish` appears only as `--dry-run` in `doc/TESTING.md` and `doc/QUICK_TEST.md` — so that step stays manual and deliberate.

Worth flagging separately: `main` carries a bot commit reading *Bumped to Version 1.4.0*, but there is no `v1.4.0` tag or release, and pub.dev's newest stable is **1.2.0** — so 1.3.0 and 1.4.0 were never published. Found while tracing the release history; neither caused nor fixed here.

## If you would rather releases stay manual for now

Drop `develop` from `release.yml`'s `on.push.branches`, keeping it in `.releaserc`. A maintainer then runs the Release workflow against `develop` when a beta is wanted, and semantic-release still works out that it is `beta.4`, `beta.5` and so on — the hand-versioning goes away without a beta appearing on every merge. Say the word and I will push that as a one-line change.

## Verification

CI configuration, so none of it is device-verifiable. What was done:

- `npx semantic-release --dry-run --no-ci` on a `develop` checkout, before and after the fix, with a probe commit to confirm the computed version (table above)
- `update-version.py` exercised on `2.1.0`, `2.0.0-beta.4`, `2.0.0-rc.1`, `3.0.0+7`, and end to end against both pubspecs and the lockfile: a bump changes exactly one lockfile line and leaves the following package untouched; a repeat run is a no-op; an absent entry and an absent file both return `False`
- `.releaserc` validated as JSON; commit messages pass the `semantic-commits` workflow's own regex
- No Dart source is touched by this PR


deepak-dhwani (approved):
Approving. I verified the load-bearing claims against semantic-release 24.2.9's own source rather than taking the description's word for them:

- `lib/utils.js:81` — `isSameChannel(c, o) => c === o || (!c && !o)`, and `lib/branches/normalize.js`'s prerelease normalisation keeps `channel: false` as `false` (`isNil(channel) ? name : channel`). So `isSameChannel(false, null)` is **true** and the hand-cut tags become visible; without the flag the channel is `"develop"` and they are not. ✅
- `lib/get-last-release.js` — prerelease tags are channel-filtered, but stable tags always pass (`|| !semver.prerelease(tag.version)`). That is precisely why the fallback lands on `v1.3.0` rather than "no previous release", which is the detail that makes this a silent downgrade instead of a loud failure. ✅
- `git notes --ref semantic-release` on all 8 tags: **empty on every one**, stables included. Diagnosis confirmed. ✅
- `lib/get-next-version.js` — confirms `2.0.0-beta.4`. ✅
- The newer CHANGELOG claim that automated releases write `{"channels":[null]}` themselves also checks out: `index.js:176` sets `channel: context.branch.channel || null`, and `index.js:209` writes `addNote({ channels: [nextRelease.channel] })`. So hand-cut and automated tags stay mutually visible going forward — this isn't just a patch over the three tags that exist today.

No changes requested. Two things are follow-ups rather than blockers, but I'd rather they were written down than discovered:

**1. The stable side is the landmine this leaves armed.** At the current `develop` there are 245 commits in `v1.3.0..develop` — 97 `fix`, 21 `feat`, and **zero** carrying a `BREAKING CHANGE:` footer or a `!` marker. Since `get-last-release` excludes prerelease tags on a *release* branch, the first `develop` → `main` merge will resolve last-release `v1.3.0`, compute `minor`, and cut **`v1.4.0`** — below the `2.0.0-beta.3` already on pub.dev. `EINVALIDNEXTVERSION` won't catch it either, because 1.4.0 > 1.3.0 is a legal forward step. Worth deciding how `2.0.0` gets cut — hand-tag, a commit with a `BREAKING CHANGE:` footer, or backfilled notes — before that merge rather than at it.

**2. Each beta will cost a doubled CI run.** `@semantic-release/git` pushes the bump commit with the GitHub App installation token, and App-token pushes *do* trigger workflows — so every release re-fires `ci.yml` (~4m30s) and `release.yml` on `develop`. No loop, since `chore` is `"release": false` and the run exits "no release published", but it is new cost that today only lands on rare `main` pushes.

Because of (2) and the cadence — and since publishing to pub.dev stays manual anyway — I'd take the alternative you offered at the bottom of the description: keep `develop` in `.releaserc`, drop it from `release.yml`'s `on.push.branches`, and cut betas via `workflow_dispatch`. That still removes the hand-versioning, which was the actual goal. Your call, not a condition of approval.

Suggest merging this ahead of #116, so the pipeline is armed when those `fix:` commits land and beta.4 cuts cleanly.

ankitDhwani:
@deepak-dhwani taken — `a274820` implements your recommendation, which re-requests review since it lands after your approval. It is one trigger line plus comments; no logic changed.

```diff
 on:
   push:
-    branches: [ main, master, develop ]
+    branches: [ main, master ]
   workflow_dispatch:
```

`develop` **stays** in `.releaserc` `branches` — semantic-release refuses to release from a branch it was not configured for, so it has to remain there for a dispatch against `develop` to work at all. What changes is that merging to `develop` now releases nothing, and a beta is fired deliberately.

Your doubled-CI-run point is the one that decided it. I had checked the loop was safe (the re-triggered run exits "no release published") but only costed it on `main`, where releases are rare — I had not costed it at one-per-beta on `develop`. Combined with pub.dev publishing being manual regardless, an automatic tag buys nothing a consumer can see and costs ~9 minutes of runner time per beta.

Both reasons are now in the workflow's comment block, along with the one-line path back to automatic (`develop` into `push.branches`), so the next person does not have to re-derive why it is absent.

The CHANGELOG entry is updated to match — it previously described the push trigger.

Also worth recording your two follow-ups here rather than losing them in a thread:

1. **The `main` 1.4.0 landmine stands and this PR does not address it.** 245 commits in `v1.3.0..develop`, zero with a `BREAKING CHANGE:` footer, so the first `develop` → `main` merge computes `minor` from `v1.3.0` and cuts `v1.4.0` — below the `2.0.0-beta.3` already on pub.dev, and `EINVALIDNEXTVERSION` will not catch it because 1.4.0 > 1.3.0 is a legal step. Worth deciding the route to `2.0.0` (hand-tag, a `BREAKING CHANGE:` footer, or backfilled notes) before that merge rather than at it. Happy to open an issue if you would rather it were tracked than sitting in this thread.

2. Noted on merge order — #117 before #116. `CHANGELOG.md` is the only overlap, so #116 will need a one-hunk rebase after this lands; I will do that rather than leave it to you.

Omprakash-48 (approved):

deepak-dhwani:
Re-reviewed at `a274820` — this is exactly the change, and the comment block is better than the one-liner I asked for: it records *why* `develop` stays in `.releaserc` while being absent from `push.branches`, which is the non-obvious half and the thing a future maintainer would otherwise "fix" by deleting.

Verified nothing else moved: `.releaserc` is untouched, so the `channel: false` reasoning I checked against semantic-release 24.2.9's source still holds, and `develop` remains a configured prerelease branch — `workflow_dispatch` against it will still resolve `v2.0.0-beta.3` and compute `beta.4`. My approval stands.

Worth restating the one open item so it doesn't get lost now that this is manual: the **stable** side is still unarmed. 245 commits in `v1.3.0..develop`, zero with a `BREAKING CHANGE:` footer or `!` marker, and `get-last-release` excludes prerelease tags on a release branch — so a push to `main` computes `minor` from `v1.3.0` and cuts `v1.4.0`, below the `2.0.0-beta.3` already on pub.dev, with no guard tripping. That is unchanged by this PR and not caused by it, but `main` **is** still on the push trigger, so it is reachable without anyone choosing to cut a release.

